PR #2, opened 2020-09-14
Fix multicore for K210 platform.

wyfcyx:
* decrease RustSBI size from 2M to 128K;
* shrink SBI heap and hart stack size to 32K and 16K respectively;
* wakeup hart1 after initialization so that all harts can go into S Mode freely.


luojia65:
可以

PR #4, opened 2020-10-12
Fix devintr for k210 platform

wyfcyx:
Add a new sbi call specially for K210 platform in order to fit its flaw of S mode external interrupt and make the device handlers on K210 and Qemu platform look the same.
|Function Name|Function ID|ExtensionID|Replacement Extension|
|-|-|-|-|
|sbi_register_devintr|0|0x09|N/A|
## Interface in C
```c
struct sbiret sbi_register_devintr(uint64_t devintr);
```
Here `devintr` is **physical address** of the external interrupt handler which is defined and implemented in S mode by yourself, it should be saved in RustSBI during this ecall. After that, the corresponded external interrupt handler will be functionally called by RustSBI when a M mode external interrupt is triggered. Interestingly, the IRQ handler manipulates S mode data structures while it executes in M mode throughout. In the IRQ handler you should provide a full claim/complete session, for example:
``` rust
/// It will be executed in M mode, and it can access kernel address space
/// after setting mstatus.mprv.
pub unsafe fn devintr() {
    // on k210, we only allow M mode devintr to be received on
    // hart0-M target after configuring PLIC.
    let hart0m_claim = 0x0C20_0004 as *mut u32;
    let irq = hart0m_claim.read_volatile();
    match irq {
        33 => {
            // UARTHS
            let mut c = (0x3800_0004 as *const u32).read_volatile();
            if c <= 255 {
                STDIN.push(c as u8);
            }
        }
        _ => {
            panic!("unsupported device interrupt!");
        }
    }
    hart0m_claim.write_volatile(irq);
}
```
## Limitations
Current implementation of RustSBI relys on memory privilege mechanics of RISC-V, leading to some limitations:
* Input argument `devintr` must be a physical address;
* Writer of the IRQ hander from S mode application should take the responsibility to guarantee that the page table before trapping into M mode is able to access the whole physical memory via virtual address which is equally mapped to physical address.
 
I have not decided what `sbiret` it returns yet since it's not vital.

## Extra Resources
[Here](https://github.com/wyfcyx/osnotes/blob/master/book/%E5%BD%BB%E5%BA%95%E8%A7%A3%E5%86%B3k210%E5%A4%96%E9%83%A8%E4%B8%AD%E6%96%AD.md) is a Chinese version report that shows how I solve this problem with some detailed code changes in RustSBI, and [here](https://github.com/wyfcyx/rCore-Tutorial/commit/2651e18973cf0b4dab68fe4a90a30a2cde7a7593) is the example project based on modified RustSBI.

### DIFF
--- a/platform/k210/src/main.rs
+++ b/platform/k210/src/main.rs
@@ -23,6 +23,8 @@ use riscv::register::{
 #[global_allocator]
 static ALLOCATOR: LockedHeap = LockedHeap::empty();
 
+static mut DEVINTRENTRY: usize = 0;
+
 #[cfg(not(test))]
 #[panic_handler]
 fn panic(info: &PanicInfo) -> ! {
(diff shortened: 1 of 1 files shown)


luojia65:
Nice work!

PR #5, opened 2020-10-16
Disabling M external when software delegating M timer into S mode

wyfcyx:
In M timer handler of RustSBI, we just disable M timer and software delegate it to S timer by setting mip.stip. Later the S application can call `set_timer` to update mtimecmp, enable M timer, and clear mip.stip(actually it should be cleared by S application). They are start and end of the timer handle session.
However, after we introduced `sbi_register_devintr`, a deadlock problem occured. To be specific, when we are in S timer handler delegated by M timer before calling `set_timer`, a M external may be triggered and handled. On K210 platform, we know it will call `devintr` to manipulate S mode data structures while still in M mode. If we acquire a lock in S timer handler and acquire it in `devintr` again before we releasing it(it can happen for the reason that S timer handler and `devintr` both access S mode data), we know it leads to a deadlock issue.
The simplest solution is disable M external in M timer handler and enable it in `set_timer` just like M timer, forbidding M external to be triggered during the timer handle session. See `platform/k210/src/main.rs` and ` rustsbi/src/ecall/legacy.rs`.

### DIFF
--- a/platform/k210/src/main.rs
+++ b/platform/k210/src/main.rs
@@ -365,6 +365,14 @@ extern "C" fn start_trap_rust(trap_frame: &mut TrapFrame) {
             if trap_frame.a7 == 0x09 {
                 unsafe { DEVINTRENTRY = trap_frame.a0; }
             } else {
+                if trap_frame.a7 == 0x0 {
+                    unsafe {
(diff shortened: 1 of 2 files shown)


wyfcyx:
RustSBI ecall global modification -> k210 local

PR #7, opened 2020-12-11
Allow legacy console_getchar work again on k210 platform.

wyfcyx:
In the previous version on k210 platform, users have to register a device interrupt handler to deal with the keyboard input manually and clear the corresponding interrupt, otherwise rustsbi will jump to a default entry(0x0) after receiving the external interrupt, leading lots of issues and finally causing rustsbi panic. However, for users who do not care performance and just want to busy loop, the conventional `console_getchar` interface does not work because of the same reason.

Now, if you do not call `sbi_rustsbi_k210_sext`, the `mie.mext` will never be set. Thus users can choose freely between calling `console_getchar` and handle all external interrupts by themselves.

Additionally, on qemu platform, LoadFault/InstructionFault/StoreFault are also delegated to S mode to handle for some reasons.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
 /target
 /.idea
+/platform/k210/.idea
+/platform/qemu/.idea
+/rustsbi/.idea
 Cargo.lock
 
 .DS_Store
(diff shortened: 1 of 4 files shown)


luojia65:
Nice! Thank you.

PR #20, opened 2021-08-16
fix: Use \r\n in println instead of \n

duskmoon314:
We noticed that we need `\r\n` instead of `\n`

For `println!("Hello"); println!("World");`

`\n`
```
Hello
     World
```

`\r\n`
```
Hello
World
```

May need more test


luojia65:
Nice! Will check it soon

PR #23, opened 2021-12-29
style: use crlf instead of lf in logo

duskmoon314:
In some systems, cr is used to indicate a line break.
Better generality can be obtained by using crlf.


luojia65:
Nice!

PR #25, opened 2022-02-12
rustsbi: guest instance; u32 function and module id

luojia65:
Instance based interface could make it convenient to develop a hypervisor. That requires a guest pointer width different to host pointer width. This feature is gated under 'guest' feature gate; current bootloader design does not need to change code for default feature of RustSBI package does not contain 'guest' feature.

This pull request also includes a u32-indexed modules and functions to comply with RISC-V SBI version 1.0. This is only an internal design change, the current API remains usize-indexed to reduce the cost of development when handling trap contexts (you don't need to change them to type u32).

Draft: instance based interface has not (by now) been proved under concept; will be marked as ready after a hypervisor with integrated RustSBI is developed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.2.0] - 2021-02-23
 ### Added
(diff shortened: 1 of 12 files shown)


luojia65:
Will land in 0.2.1 version

PR #27, opened 2022-05-16
feat: support 2021 inline variable capture style for `println!`

YdrMaster:


luojia65:
Thanks! It would allow to use 2021 inline variable and provide reference to relevant kernel projects.

PR #28, opened 2022-05-17
fix: `hart_stop` can only stop the calling hart

YdrMaster:
In SBI 1.0.0 Spec, `sbi_hart_stop` is described as:

> 9.2. Function: HART stop (FID #1)
>
> ```c
> struct sbiret sbi_hart_stop(void)
> ```
>
> Request the SBI implementation to stop executing the calling hart in supervisor-mode and return it’s ownership to the SBI implementation. This call is not expected to return under normal conditions. The sbi_hart_stop() must be called with the supervisor-mode interrupts disabled.


luojia65:
Thanks! This important fix should be included in 0.3.0 release.

PR #37, opened 2022-08-08
fix: marchid, mimpid registers in base extension

luojia65:
upstreams https://github.com/rust-embedded/riscv/pull/107. We should change to specific release number in the future other than a git commit number.


orangecms (approved):
🥳 that was a hard thing to find 😅

YdrMaster (approved):

luojia65:
Updated to release number 0.9.0 in https://github.com/rustsbi/rustsbi/commit/7370590cd4910eb1bcfecf437d71081503567bb2

PR #39, opened 2022-08-10
ecall: wrap mip::clear_ssoft() in unsafe

orangecms:
Clearing a CSR is marked as being unsafe:
https://github.com/rust-embedded/riscv/blob/e38a68d2e/src/register/macros.rs#L185

Signed-off-by: Daniel Maslowski <info@orangecms.org>


luojia65 (approved):

luojia65:
Thanks! :)

PR #47, opened 2023-04-15
ci: add dependabot for cargo and github action

zjx1319:
This PR add the dependabot configuration file `.github/dependabot.yml` to enable daily version updates for dependencies. It will ensure that our dependencies are regularly checked for new versions and updated as necessary.


luojia65:
Thanks! :)

PR #48, opened 2023-04-23
feat: support System Suspend extension and CPPC extension

jakezhu9:
Add support for System Suspend extension and CPPC extension. 

Renference: 
https://github.com/riscv-non-isa/riscv-sbi-doc/blob/master/riscv-sbi.adoc#sbi-system-suspend-extension-eid-0x53555350-susp
https://github.com/riscv-non-isa/riscv-sbi-doc/blob/master/riscv-sbi.adoc#cppc-extension-eid-0x43505043-cppc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


luojia65:
Looks good. Next step we should implement those two extension as well in instance.rs file.

PR #50, opened 2023-04-28
Create LICENSE.txt

ultraman12138:
MULAN_LICENSE is available now after modifying


luojia65:
We already have LICENSE-MULAN file in project root; please describe what would be the benefit of adding a new identical file.

PR #58, opened 2023-12-07
Add support for nacl and sta extension

jakezhu9:


luojia65:
Nice! We should also write in `instance` module to include NACL and STA extensions.

PR #59, opened 2023-12-20
Derive macro based RustSBI implementation

luojia65:
This patch replaces instance based RustSBI with `#[derive(RustSBI)]`. Now we can write:

```rust
#[derive(RustSBI)]
struct MySBI {
    console: MyConsole,
    hsm: MyHsm,
    // ...
}
```

... and RustSBI automatically derives trait `RustSBI`, i.e. the function `handle_ecall` for us. Now we can call `sbi.handle_ecall(...)` in trap handler to use the derive code as a minimum cost wrapper of implementing RISC-V SBI.


luojia65:
Note: content in this pull request is discussed with @YdrMaster and @duskmoon314 . Thanks!

PR #68, opened 2024-05-08
fix: derive dynamic RustSBI on struct with generics(type, const, lifetime, where clause)

Plucky923:
In PR #67 , a RustSBI macro implementation that can be processed dynamically has been added. 

But the current implementation has some problems. Take the following struct as an example.
``` rust
#[derive(RustSBI)]
#[rustsbi(dynamic)]
struct WithLifetime<'a> {
    info: &'a DummyEnvInfo,
}
```
The macro code cannot compile this struct because its `strcut prober` struct does not handle lifetime situations. In fact, all generics are not handled, including lifetime, type generics, constant generics, and where clauses. These problems limit the application of `rustsbi dynamic` in actual projects.

This pull request solves the problem of generics not being processed. The definition of `struct _Prober` is extended to include generics, which is composed of impl generics, type generics and where clause. We add them to the generated code of `struct _Prober`, thus solving the problem.
``` rust
struct _Prober #impl_generics (&'_lt #name #origin_ty_generics) #where_clause;
impl #impl_generics ::rustsbi::_ExtensionProbe for _Prober #ty_generics #where_clause 
```
Until now,`rustsbi dynamic` can be used for struct with lifetime, and its application scope will be expanded.

In order to verify the effectiveness of this modification, an additional unit test module(`tests/dynamic-generics.rs`) has been added. All structures in this unit test file contain different types of generics to comprehensively test the correctness of the macro generated code, thereby verifying the repair results of this pull request.

### DIFF
--- a/macros/Cargo.toml
+++ b/macros/Cargo.toml
@@ -21,6 +21,7 @@ proc-macro = true
 [dependencies]
 quote = "1.0.33"
 syn = "2.0.39"
+proc-macro2 = "1.0.82"
 
 [features]
 default = []
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -3,8 +3,12 @@
 //! Do not use this crate directly.
 
 use proc_macro::TokenStream;
+use proc_macro2::Span;
 use quote::{quote, ToTokens};
-use syn::{parse_macro_input, Data, DeriveInput, Generics, Ident, Member};
+use syn::{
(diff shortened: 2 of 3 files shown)


YdrMaster (approved):

luojia65 (approved):
Good job, the added unit test file is great as well.

duskmoon314 (approved):
LGTM

PR #69, opened 2024-05-08
fix: implement all rustsbi traits for `Option<T>`

Placebo27:
Add `Option<T>` implementation for `console`, `cppc`, `hsm`, `ipi`, `nacl`, `pmu`, `reset`, `sta`, `susp`, `timer` traits.
Assist rustsbi dynamic feature. Now we can use the `Option<T>` type in `#[rustsbi(dynamic)]` structs, extending the application scope of the rustsbi dynamic function.

### DIFF
--- a/src/console.rs
+++ b/src/console.rs
@@ -106,3 +106,31 @@ impl<T: Console> Console for &T {
         T::write_byte(self, byte)
     }
 }
+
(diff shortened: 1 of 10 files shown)


luojia65 (approved):
Nice catch!

PR #71, opened 2024-05-21
spec: convert `BitVector` to a helper function, add counter index mask

rmsyn:
Changes the `BitVector` struct to a trait, and implements the trait for `HartMask`.

Allows for code re-use of the `BitVector` functions without wrapping the type.

Adds the `CounterIndexMask` type to represent `counter_idx_mask` and `counter_idx_base` parameters from the `PMU` extension.

### DIFF
--- a/sbi-spec/src/binary.rs
+++ b/sbi-spec/src/binary.rs
@@ -742,6 +742,51 @@ impl HartMask {
     }
 }
 
+/// Counter index mask structure in SBI function calls for the `PMU` extension §11.
+#[repr(C)]
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub struct CounterMask {
+    counter_idx_mask: usize,
+    counter_idx_base: usize,
+}
+
+impl CounterMask {
+    /// Special value to ignore the `mask`, and consider all `bit`s as set.
+    pub const IGNORE_MASK: usize = usize::MAX;
+
+    /// Construct a [CounterMask] from mask value and base counter index.
+    #[inline]
+    pub const fn from_mask_base(counter_idx_mask: usize, counter_idx_base: usize) -> Self {
+        Self {
+            counter_idx_mask,
+            counter_idx_base,
+        }
+    }
+
+    /// Gets the special value for ignoring the `mask` parameter.
+    #[inline]
+    pub const fn ignore_mask(&self) -> usize {
+        Self::IGNORE_MASK
+    }
+
+    /// Returns `mask` and `base` parameters from the [CounterMask].
+    #[inline]
+    pub const fn into_inner(self) -> (usize, usize) {
+        (self.counter_idx_mask, self.counter_idx_base)
+    }
+
+    /// Returns whether the [CounterMask] contains the provided `counter`.
+    #[inline]
+    pub const fn has_bit(self, counter: usize) -> bool {
+        has_bit(
+            self.counter_idx_mask,
+            self.counter_idx_base,
+            Self::IGNORE_MASK,
+            counter,
+        )
+    }
+}
+
 /// Physical slice wrapper with type annotation.
 ///
 /// This struct wraps slices in RISC-V physical memory by low and high part of the
@@ -917,6 +962,36 @@ mod tests {
         for i in 0..5 {
             assert!(mask.has_bit(i));
         }
+        assert!(mask.has_bit(usize::MAX));
+    }
+
+    #[test]
+    fn rustsbi_counter_index_mask() {
+        let mask = CounterMask::from_mask_base(0b1, 400);
+        assert!(!mask.has_bit(0));
+        assert!(mask.has_bit(400));
+        assert!(!mask.has_bit(401));
+        let mask = CounterMask::from_mask_base(0b110, 500);
+        assert!(!mask.has_bit(0));
+        assert!(!mask.has_bit(500));
+        assert!(mask.has_bit(501));
+        assert!(mask.has_bit(502));
+        assert!(!mask.has_bit(500 + (usize::BITS as usize)));
+        let max_bit = 1 << (usize::BITS - 1);
+        let mask = CounterMask::from_mask_base(max_bit, 600);
+        assert!(mask.has_bit(600 + (usize::BITS as usize) - 1));
+        assert!(!mask.has_bit(600 + (usize::BITS as usize)));
+        let mask = CounterMask::from_mask_base(0b11, usize::MAX - 1);
+        assert!(!mask.has_bit(usize::MAX - 2));
+        assert!(mask.has_bit(usize::MAX - 1));
+        assert!(mask.has_bit(usize::MAX));
+        assert!(!mask.has_bit(0));
+        let mask = CounterMask::from_mask_base(0, usize::MAX);
+        let null_mask = CounterMask::from_mask_base(0, 0);
+        (0..=usize::BITS as usize).for_each(|i| {
+            assert!(mask.has_bit(i));
+            assert!(!null_mask.has_bit(i));
+        });
         assert!(mask.has_bit(usize::MAX));
     }
 }
--- a/sbi-spec/src/binary.rs
+++ b/sbi-spec/src/binary.rs
@@ -673,58 +673,73 @@ impl SbiRet {
     }
 }
 
+/// Check if the implementation contains the provided `bit`.
+///
+/// ## Parameters
+///
+/// - `mask`: bitmask defining the range of bits.
+/// - `base`: the starting bit index. (default: `0`)
+/// - `ignore`: if `base` is equal to this value, ignore the `mask` parameter, and consider all `bit`s set.
+/// - `bit`: the bit index to check for membership in the `mask`.
+pub(crate) const fn has_bit(mask: usize, base: usize, ignore: usize, bit: usize) -> bool {
+    if base == ignore {
+        // ignore the `mask`, consider all `bit`s as set.
+        true
+    } else if bit < base {
+        // invalid index, under minimum range.
+        false
+    } else if (bit - base) >= usize::BITS as usize {
+        // invalid index, over max range.
+        false
+    } else {
+        // index is in range, check if it is set in the mask.
+        mask & (1 << (bit - base)) != 0
+    }
+}
+
 /// Hart mask structure in SBI function calls.
-#[derive(Debug, Copy, Clone)]
+#[repr(C)]
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub struct HartMask {
-    inner: BitVector,
+    hart_mask: usize,
+    hart_mask_base: usize,
 }
 
 impl HartMask {
-    /// Construct a hart mask from mask value and base hart id.
-    #[inline]
-    pub const fn from_mask_base(hart_mask: usize, hart_mask_base: usize) -> HartMask {
-        HartMask {
-            inner: BitVector {
-                hart_mask,
-                hart_mask_base,
-            },
-        }
-    }
+    /// Special value to ignore the `mask`, and consider all `bit`s as set.
+    pub const IGNORE_MASK: usize = usize::MAX;
 
-    /// Returns `hart_mask` and `hart_mask_base` parameters from the hart mask structure.
+    /// Construct a [HartMask] from mask value and base hart id.
     #[inline]
-    pub const fn into_inner(self) -> (usize, usize) {
-        (self.inner.hart_mask, self.inner.hart_mask_base)
-    }
-
-    /// Check if the `hart_id` is included in this hart mask structure.
-    #[inline]
-    pub const fn has_bit(&self, hart_id: usize) -> bool {
-        let BitVector {
+    pub const fn from_mask_base(hart_mask: usize, hart_mask_base: usize) -> Self {
+        Self {
             hart_mask,
             hart_mask_base,
-        } = self.inner;
-        if hart_mask_base == usize::MAX {
-            // If `hart_mask_base` equals `usize::MAX`, that means `hart_mask` is ignored
-            // and all available harts must be considered.
-            return true;
         }
-        let Some(idx) = hart_id.checked_sub(hart_mask_base) else {
-            // hart_id < hart_mask_base, not in current mask range
-            return false;
-        };
-        if idx >= usize::BITS as usize {
-            // hart_idx >= hart_mask_base + XLEN, not in current mask range
-            return false;
-        }
-        hart_mask & (1 << idx) != 0
     }
-}
 
-#[derive(Debug, Copy, Clone)]
-struct BitVector {
-    hart_mask: usize,
-    hart_mask_base: usize,
+    /// Gets the special value for ignoring the `mask` parameter.
+    #[inline]
+    pub const fn ignore_mask(&self) -> usize {
+        Self::IGNORE_MASK
+    }
+
+    /// Returns `mask` and `base` parameters from the [HartMask].
+    #[inline]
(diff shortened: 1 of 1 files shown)


luojia65 (changes requested):
Thanks! Your pull request generally looks nice, it simplifies internal `sbi-spec` implementation of `HartMask` structure, and creates a good unification of mask structures by the `has_bit` function. The structure this pull request introduces is correctly tested using the added test case. There are some small concerns, though, please allow me to describe in detail.

## 1. Struct name of `CounterIndexMask`

The original SBI standard document defined two bit-vector based mask structure pairs, we can make a summary in the following table.

| Name of Mask | Name of Base | Reference |
|---|---|---|
| `hart_mask` | `hart_mask_base` | [IPI Extension](https://github.com/riscv-non-isa/riscv-sbi-doc/blob/9f98c858d2b17506d0ef8b13696ac428a258f800/src/ext-ipi.adoc#function-send-ipi-fid-0) |
| `counter_idx_mask` | `counter_idx_base` | [PMU Extension](https://github.com/riscv-non-isa/riscv-sbi-doc/blob/9f98c858d2b17506d0ef8b13696ac428a258f800/src/ext-pmu.adoc#function-find-and-configure-a-matching-counter-fid-2) |

As we can see from this table, the RISC-V SBI specification doesn't unify the naming style of different masks and bases. For example, unlike what masks of harts are named, it does not name mask of counters like `counter_mask` and `counter_mask_base`. That makes sense for the name `CounterIndexMask` (defined [here](https://github.com/rustsbi/rustsbi/pull/71/files#diff-6a6ad5fcfbc094c6b610c783bfa81ad2371c250851741bdfaeb3203a6b5049bfR724)). But I'd like to name it `CounterMask` instead, for the following reasons:

- Both harts and counters are identified by a number, namely hart ID or counter index. Since the existing code did not name like `HartIdMask`, we may consider the `Index` redundant.
- `CounterIndexMask` is possibly too long. It's especially obvious when the user uses it with full name in function parameter like:
```rust
// original
fn function_name(counter_index_mask: rustsbi::CounterIndexMask)
// proposed
fn function_name(counter_mask: rustsbi::CounterMask)
```
Users may find out a shorter name convenient to read and remember; it also affects how the library level rustdoc is shown.

Based on the discussions above, will it look better if we name this structure `CounterMask` instead?

## 2. `BitVector` constraints, return type and usage of its functions.

This pull request introduces a new trait `BitVector` for an abstraction of all masks with `mask` and `base` as parameters. It introduces five functions, whose signatures can be described below:

```rust
pub trait BitVector {
    fn mask(&self) -> usize;
    fn base(&self) -> usize;
    fn ignore_mask(&self) -> usize;
    fn to_inner(&self) -> (usize, usize) { ... }
    fn has_bit(&self, bit: usize) -> bool { ... }
}
```

In this way we unify `has_bit` function which is convenient for code reuse of all bit vectors.

First, since bit vectors wrapping `mask` and `base` are relatively small in memory size, we can use [copy semantics](https://medium.com/coderhack-com/using-rust-variables-with-move-move-semantics-in-detail-with-examples-9f2fa71859e7#copy-types) to describe them in Rust. I didn't find out a complete description of the difference of copy or move semantics in a quick web search, but I'd like to conclude it in following aspects:

- `Copy` types are relatively small in memory size, and typically does not own any heap memory;
- We save a `.clone()` function call to copy a structure implementing `Copy`; when it's small in memory size, the structure is typically saved on registers on high optimization level (which is common on firmware development), making the `.clone()` often effectively equals to no-ops;

Since `BitVector` implementations typically include two `usize`'s only, we can assume it `Copy`able. In this way, it will be better when all its functions require `self` other than `&self`; and since this function consumes `self`, it will be better to name it with prefix `into_` other than `to_`.

Secondly, the `ignore_mask` is defined as a function. Will it be better if it is defined as an [associated const item](https://doc.rust-lang.org/reference/items/associated-items.html#associated-constants-examples) for trait `BitVector`? In this way, we align with standard library coding styles (like `usize::MAX`), and further optimizations can be made since it's a compile-time known `const`.

After the discussions above, I'd like to change definition of `BitVector` trait into:

```rust
pub trait BitVector: Copy {
    const IGNORE_MASK: usize;
    fn mask(self) -> usize;
    fn base(self) -> usize;
    fn into_inner(self) -> (usize, usize) { ... }
    fn has_bit(self, bit: usize) -> bool { ... }
}
```

When we document `BitVector` trait, we should describe why we expose the abstraction of `BitVector` to user level. In my opinion, when we are introducing a new trait, we should take care when considering following aspects:

- Will it be used like `fn function<T: BitVector>(param: T) -> RetType`?
- Does it make sense when we define `struct StructName<T: BitVector> { inner: T }`?
- Is it acceptible when users have to `use rustsbi::BitVector;` at all times when using its functions?
- What if the downstream type overrided `into_inner` or `has_bit` functions?

If it's not the case (is this abstraction confusing or not that useful on downstream software?), we could treat it as an internal implementation by not making this trait public. In this way, the `has_bit` function can be defined as an internal function, other than an associated function of `BitVector`.

```rust
// alternatively, make it private
trait BitVector { ... }
impl HartMask {
    pub const fn from_mask_base(hart_mask: usize, hart_mask_base: usize) -> HartMask { ... }
    // into_inner is in trait `BitVector`
}
impl CounterIndexMask {
    pub const fn from_mask_base(hart_mask: usize, hart_mask_base: usize) -> CounterIndexMask { ... }
    // into_inner is in trait `BitVector`
}
// ... or, remove this trait and replace it using `has_bit` internal helper function
fn has_bit(mask: usize, base: usize, bit: usize) -> bool { ... }
impl HartMask {
    pub const fn from_mask_base(hart_mask: usize, hart_mask_base: usize) -> HartMask { ... }
    pub const fn into_inner(self) -> (usize, usize) { ... }
    pub const fn has_bit(bit: usize) -> bool { ... } // uses internal has_bit function
}
impl CounterIndexMask {
    pub const fn from_mask_base(hart_mask: usize, hart_mask_base: usize) -> CounterIndexMask { ... }
    pub const fn into_inner(self) -> (usize, usize) { ... }
    pub const fn has_bit(bit: usize) -> bool { ... } // uses internal has_bit function
}
```

@rmsyn Can we add further modifications to this pull request considering the discussions above? Thanks!

rmsyn:
> But I'd like to name it CounterMask instead

That sounds good to me, I can include the name change in the set of fixup changes.

> Copy types are relatively small in memory size, and typically does not own any heap memory

I think this change, and the `self` parameters are going to also require a `Sized` constraint on the trait. I'll experiment with some changes to see what works.

I tried to implement the `into_inner(self) -> (usize, usize)`, and the compiler complained about the size of the type being unknown.

> When we document BitVector trait, we should describe why we expose the abstraction of BitVector to user level.

I agree, I can add some more documentation about types that should implement `BitVector`.

> or, remove this trait and replace it using `has_bit` internal helper function

I think `BitVector` or the associated helper `has_bit` would be generally useful outside the crate. For example, any place where a user has a bitmask and base bitfield offset, they could use `has_bit` to check a bitfield.

Personally, I'd prefer the trait. Mostly because we can enforce more invariants, and help provide more proper use guidance.

> from_mask_base(hart_mask: usize, hart_mask_base: usize)

What are your thoughts on also introducing two new types `struct BitMask(usize)` and `struct BitBase(usize)`, and changing the above signature to:

```rust
fn from_mask_base(mask: BitMask, base: BitBase) -> Self { ... }
```

This would help the user from swapping the parameters somewhat, though I'm not sure if the extra constraint is worth the effort (maybe over-engineering).

rmsyn:
> When we document BitVector trait, we should describe why we expose the abstraction of BitVector to user level.

After applying some of the suggestions you made, I think the helper function is probably the better solution. Requiring users to import the `BitVector` trait is really too much for how simple the functionality is.

If there ends up being a number of similar types added to the SBI spec, we could also define a helper macro to create these `*Mask` types. With just the two current types, its probably not worth the macro just yet.

> We save a .clone() function call to copy a structure implementing Copy

Was this just in reference to the `into_inner` versus `to_inner` discussion?

I've added the `ignore_mask` function with a `&self` receiver, however this function neither copies nor clones the type. It's just a convenience function to get the `IGNORE_MASK` value from a concrete instance, without having to do something like:

```rust
HartMask::IGNORE_MASK
```

luojia65:
@rmsyn: Thanks for your contribution! I'm okay with the new `CounterMask` structure, internal improvements of `has_bit` function, and the test cases added. The usage of associated `IGNORE_MASK` constant and `ignore_mask` function (in my opinion only) should be furtherly discussed, but since `sbi-spec` crate is in pre-0.1 stage by version, I'm okay to approve it first for the community to test it out.

r? @YdrMaster @duskmoon314

luojia65 (approved):
@rmsyn: Thanks for your contribution! I'm okay with the new `CounterMask` structure, internal improvements of `has_bit` function, and the test cases added. The usage of associated `IGNORE_MASK` constant and `ignore_mask` function (in my opinion only) should be furtherly discussed, but since `sbi-spec` crate is in pre-0.1 stage by version, I'm okay to approve it first for the community to test it out.

> What are your thoughts on also introducing two new types struct BitMask(usize) and struct BitBase(usize), and changing the above signature to:
> fn from_mask_base(mask: BitMask, base: BitBase) -> Self { ... }

In my opinion, this design may result in a rise of learning curve for new users. It requires users always have to wrap the mask and base parameters in `BitMask` and `BitBase` structures, sometimes it may be redundant or harder to use. Some users may suggest to change parameter into `impl Into<BitMask>` etc. to improve easiness on usage, but since (in my thoughts only) `usize` would have to implement both `Into<BitMask>` and `Into<BitBase>`, this design won't have the effect to prevent accidentially swap of both parameters.

There are some opinions on similiar ways of preventing wrong order of parameters, like in the [`remote_hfence_vvma_asid`](https://docs.rs/sbi-rt/0.0.3/sbi_rt/fn.remote_hfence_vvma_asid.html) function of the `sbi-rt` crate, there are three `usize` parameters that may be in wrong order. We can leave the issue here as-is, for further discussions on better possible solutions for this problem.

Since `rustsbi` and its ecosystem crates in the rustsbi/rustsbi repository often resides in S-mode or M-mode security critical areas on RISC-V architecture, I'd like more of our maintainters to leave comments.

r? @YdrMaster @duskmoon314

duskmoon314:
Since there is no trait in the current implementation, I suggest renaming the PR title.

rmsyn:
> In my opinion, this design may result in a rise of learning curve for new users. It requires users always have to wrap the mask and base parameters in BitMask and BitBase structures, sometimes it may be redundant or harder to use. Some users may suggest to change parameter into impl Into<BitMask> etc. to improve easiness on usage, but since (in my thoughts only) usize would have to implement both Into<BitMask> and Into<BitBase>, this design won't have the effect to prevent accidentially swap of both parameters.

I was thinking along the same lines, that requiring the extra types doesn't really get us much. One alternative could be to adopt a "builder" type interface using something like:

```rust
struct SomeMask {
    mask: usize,
    base: usize,
}

impl SomeMask {
    pub const fn new() -> Self {
        Self { mask: 0, base: 0 }
    }
    /// This is a builder function to set the `mask` field
    pub const fn with_mask(mut self, mask: usize) -> Self {
        self.mask = mask;
        self
    }

    /// This is a builder function to set the `base` field
    pub const fn with_base(mut self, base: usize) -> Self {
        self.base = base;
        self
    }
}
```

Generally, when I create a `struct` type, I provide three types of "accessor" function:

- `getter`: same name as the field, usually has a `pub const fn field_name(&self) -> FieldType` signature
  - if the field is non-`Copy`: `pub const fn field_name(&self) -> &FieldType`  
- `setter`: same name as the field, prefixed with `set_`: `pub fn set_field_name(&mut self, val: FieldType)`
- `wither`: builder function that calls the `setter`, and returns `Self`:
```rust
pub fn with_field_name(mut self, val: FieldType) -> Self {
    self.set_field_name(val);
    self
}
```

rmsyn:
> In my opinion, this design may result in a rise of learning curve for new users. It requires users always have to wrap the mask and base parameters in BitMask and BitBase structures, sometimes it may be redundant or harder to use. Some users may suggest to change parameter into impl Into<BitMask> etc. to improve easiness on usage, but since (in my thoughts only) usize would have to implement both Into<BitMask> and Into<BitBase>, this design won't have the effect to prevent accidentially swap of both parameters.

I was thinking along the same lines, that requiring the extra types doesn't really get us much.

One alternative could be to adopt a "builder" type interface using something like:

```rust
struct SomeMask {
    mask: usize,
    base: usize,
}

impl SomeMask {
    /// Create a `default` [SomeMask].
    pub const fn new() -> Self {
        Self { mask: 0, base: 0 }
    }

    /// The current creation function.
    pub const fn from_mask_base(mask: usize, base: usize) -> Self {
        Self { mask, base }
    }

    /// This is a builder function to set the `mask` field
    pub const fn with_mask(self, mask: usize) -> Self {
        Self { mask, base: self.base }
    }

    /// This is a builder function to set the `base` field
    pub const fn with_base(self, base: usize) -> Self {
        Self { mask: self.mask, base }
    }
}

// Then creating a new mask can look like this
let some_mask = SomeMask::new().with_mask(0x3f).with_base(1);
// or a const
const SOME_MASK: SomeMask = SomeMask::new().with_mask(0x3f).with_base(1);
```

Generally, when I create a `struct` type, I provide three types of "accessor" function:

- `getter`: same name as the field, usually has a `pub const fn field_name(&self) -> FieldType` signature
  - if the field is non-`Copy`: `pub const fn field_name(&self) -> &FieldType`  
- `setter`: same name as the field, prefixed with `set_`: `pub fn set_field_name(&mut self, val: FieldType)`
- `wither`: builder function that sets the field, and returns `Self`

duskmoon314:
> * `setter`: same name as the field, prefixed with `set_`: `pub fn set_field_name(&mut self, val: FieldType)`

I personally don't like this naming convention in Rust. `foo_mut` is often used.

---

I'm not very familiar with this part. I wonder whether the mask and the base change throughout its life. If not, maybe something like `Mask<const BASE: usize, const MASK: usize>` is also usable.

rmsyn:
> `foo_mut` is often used.

I think this is usually for functions that return a mutable reference to the `foo` field.

> `Mask<const BASE: usize, const MASK: usize>`

This could work, but it is a much more unwieldy interface. For instance, most of the places where a `*Mask` type would be used are taking in parameters passed in from the `Supervisor` program.

This means it would be impossible to create a `Mask<const BASE: usize, const MASK: usize>` without a bunch of helper machinery. Where the current `struct` only requires the additional `with_base` and `with_mask` functions for clarity.

All of the creation and `with_*` functions being `const` also allow for the construction of a known constant.

So, its possible to create a `struct Mask` instance in a variable or `const` context, from variables and/or `const`s.

However, a `Mask<const BASE: usize, const MASK: usize>` can only be created from `const`s.

duskmoon314:
> I think this is usually for functions that return a mutable reference to the `foo` field.

Yes, and the usage would be something like `xxx.foo_mut() = yyy`

rmsyn:
> Yes, and the usage would be something like xxx.foo_mut() = yyy

There are downsides to interfaces like that. For instance, it creates a mutable reference to `xxx` that isn't released until the binding variable goes out of scope.

That's why the Rust `set_` convention is usually preferred, because it holds the mutable reference for the minimal amount of time.

luojia65:
If possible, we can leave the `&mut self` methods for further discussions, and always create new mask structures by now.

luojia65:
I'm merging this for that current code already satisfied the scope of this pull request. For further discussions on 'set' and 'mut' functions, we can fire new issues or pull requests on it.

YdrMaster (approved):

PR #82, opened 2024-11-30
spec: add some methods for HartMask

woshiluo:
- Add `HartIds` structure as iteration of HartMask, also add test for testing it.
- Add `HartMask::insert` and `HartMask::remove`, and add its test.

### DIFF
--- a/sbi-spec/src/binary.rs
+++ b/sbi-spec/src/binary.rs
@@ -794,6 +794,12 @@ pub struct HartMask {
     hart_mask_base: usize,
 }
 
+/// Iteration for HartMask, from low to high.
+pub struct HartMaskIter {
+    inner: HartMask,
+    visited_mask: usize,
+}
+
 impl HartMask {
     /// Special value to ignore the `mask`, and consider all `bit`s as set.
     pub const IGNORE_MASK: usize = usize::MAX;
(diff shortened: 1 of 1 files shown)


sbi-spec/src/binary.rs
```diff
@@ -794,6 +804,20 @@ pub struct HartMask {
     hart_mask_base: usize,
 }
 
+/// Iteration for HartMask, from low to high.
+pub struct HartMaskIter {
```
luojia65, line 808: Please consider derive `Clone`, `Copy`, `Debug`, `Eq`, `PartialEq` and `Hash` for this structure. We exclude `Ord` and `PartialOrd` for that this structure does not define any orderings.

sbi-spec/src/binary.rs
```diff
@@ -829,6 +853,63 @@ impl HartMask {
...
+            },
+            visited_mask: 0,
+        }
+    }
+}
+
+impl Iterator for HartMaskIter {
+    type Item = usize;
+
+    fn next(&mut self) -> Option<Self::Item> {
```
luojia65, line 901: Should we `#[inline]` this function? We can compare compilation results for functions that include or does not include inline hints to see how they differs.

sbi-spec/src/binary.rs
woshiluo: Derived, thanks for point it out.

sbi-spec/src/binary.rs
woshiluo: With inline hint, it seems even unrolling the loop of `for x in mask.iter()`. 

So I added inline hints here, thanks for this notice.

luojia65 (approved):
This pull request LGTM

PR #85, opened 2024-12-26
spec: add MPXY extension support to SBI implementation

xiaobor123:
this commit add add necessary constants for Extension ID (EID) and Function ID (FID) associated with the Message Proxy Extension (MPXY). ref-https://github.com/riscv-non-isa/riscv-sbi-doc/blob/master/src/ext-mpxy.adoc


luojia65 (approved):
LGTM

PR #104, opened 2025-02-13
feat(xtask): add `qemu` subcommand

drindr:
Add `qemu` subcommand for auto launching qemu virtual machine after compiling the rustsbi to experience.


woshiluo:
Thank you for your contribution!

However, this feature feels awkward. Directly calling QEMU with its numerous arguments is standard practice, which the xtask approach doesn't currently support. Requiring a predefined QEMU path and name also adds unnecessary complexity for both users and maintainers.

Therefore I suggest to close this pull request for now. Thanks again for your effort.

PR #113, opened 2025-02-28
fix(prototyper): specify `serde-device-tree` commit id to resolve prototyper compilation error

guttatus:
Due to the break change of `serde-device-tree#4cff697c`, the prototyper has a compilation error. Therefore, specify the commit id of serde-device-tree in `Cargo.toml` as `e7f9404f` to solve the above error.


luojia65 (approved):
LGTM

PR #126, opened 2025-03-28
fix(prototyper): move device tree out of pmu protect

woshiluo:
This pull request do the following fix:

- Move device tree out of pmu protect
- Make sure pmu section aligned with 4k


guttatus (approved):
LGTM

PR #130, opened 2025-04-06
refactor(xtask): Restructure code and enhance error message clarity

chenzongyao200127:
Simplified the `print_pmu_info` function to use separate logging for available and not available PMU extensions.
Refactored the `run` function in `xtask/src/prototyper.rs` to enhance directory preparation and configuration file setup.
Improved error handling and logging for build failures and file operations in both prototyper and test kernel tasks.

### Current Error Message Info

image

image

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -31,26 +31,47 @@ struct Cli {
 
 #[derive(Subcommand)]
 enum Cmd {
+    /// Build and configure the RustSBI Prototyper bootloader.
     Prototyper(PrototyperArg),
+    /// Build test-kernel for the RustSBI Prototyper.
     Test(TestArg),
+    /// Build bench-kernel for the RustSBI Prototyper.
     Bench(BenchArg),
 }
 
 fn main() -> ExitCode {
     let cli_args = Cli::parse();
-    logger::Logger::init(&cli_args).expect("Unable to init logger");
-
-    if let Some(code) = match cli_args.cmd {
-        Cmd::Prototyper(ref arg) => prototyper::run(arg),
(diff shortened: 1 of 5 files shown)


woshiluo (approved):
LGTM

luojia65 (approved):

xtask/src/main.rs
```diff
@@ -31,26 +31,47 @@ struct Cli {
 
 #[derive(Subcommand)]
 enum Cmd {
+    /// Build and configure the RustSBI Prototyper bootloader.
     Prototyper(PrototyperArg),
+    /// Run tests for the RustSBI Prototyper.
```
guttatus, line 36:
```suggestion
    /// Build test-kernel for the RustSBI Prototyper.
```

xtask/src/main.rs
```diff
@@ -31,26 +31,47 @@ struct Cli {
...
     Test(TestArg),
+    /// Run benchmarks for the RustSBI Prototyper.
```
guttatus, line 38:
```suggestion
    /// Build bench-kernel for the RustSBI Prototyper.
```

guttatus (changes requested):

xtask/src/main.rs
chenzongyao200127: Thanks for pointing that out, I’ll fix it in the next commit.

xtask/src/main.rs
chenzongyao200127: Thanks for pointing that out, I’ll fix it in the next commit.

guttatus (approved):
LGTM

PR #132, opened 2025-04-22
fix(sbi-testing): replace `#[naked]` with `#[unsafe(naked)]` for stable Rust compatibility

Saramanda9988:
### Description
﻿
This PR addresses compilation failures in the `sbi-testing` module by updating naked function attributes to comply with Rust's stabilization of naked functions (rust-lang/rust#134213).
﻿
### Changes
- Replaced all occurrences of `#[naked]` with `#[unsafe(naked)]` in the `sbi-testing` module
- Verified compatibility with both stable and nightly Rust toolchains
- Ensured related documentation reflects this change
﻿
### Testing
﻿
**Local verification:**
- `cargo build` (both stable (1.75.0+) and nightly(1.88.0))
- `cargo test` 
- ‘cargo clippy -p sbi-testing --all-targets’
﻿
### References
- Fixes: [Issue: #131 ]
- Rust Upstream PR: [rust-lang/rust#134213](https://github.com/rust-lang/rust/pull/134213) (Stabilize `naked_functions`)
﻿
Signed-off-by: Yuyang Cai <2074730050@qq.com>

PR #138, opened 2025-06-01
fix(prototyper): replace `#[naked]` with `#[unsafe(naked)]` usage in prototyper.

jackfiled:
As #132 replaces the `#[naked]` in the `sbi-testing` package, there are still some `#[naked]` attributes in the `prototyper` package.

So in this pull requests:

- Replace all `#[naked]` with `#[unsafe(naked)]` in the `prototyper` package.
- Remove the unnecessary unsafe blocks which wraps the `naked_asm!`.


guttatus:
LGTM

guttatus (approved):

PR #141, opened 2025-07-23
feat(prototyper): allow to specify compile target in `cargo prototyper`

luojia65:
This pull request introduces `--target` parameter to `xtask` module, allowing to build RustSBI Prototyper in user specified build targets.

Both tuple-based and json-based targets are supported. If json-based target file is passed, `xtask` will find target directory according to json file names, which matches behavior of the Rustc compiler.

Usage:

```bash
cargo prototyper --target riscv64gc-unknown-none-elf
# or
cargo prototyper --target /path/to/custom/target/riscv64gc-custom_vendor-none.json
```

Parameter `--target` defaults to constant `ARCH`, which is `riscv64gc-unknown-none-elf`.

(Please squash while merging this pull request)

### DIFF
--- a/xtask/src/prototyper.rs
+++ b/xtask/src/prototyper.rs
@@ -1,6 +1,6 @@
 use std::{
     env, fs,
-    path::PathBuf,
+    path::{Path, PathBuf},
     process::{Command, ExitStatus},
 };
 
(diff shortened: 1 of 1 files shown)


woshiluo (approved):
LGTM

guttatus (approved):
LGTM

PR #145, opened 2025-08-07
fix(prototyper): use rustc_align attribute on function.

jackfiled:
The upstream is trying to stablize the `align` attribute on the function, which leading to regression. So in the [pr](https://github.com/rust-lang/rust/pull/144080), the upstream renames the `align` to `rustc_align` to fix the regression.

Fix the toolchain to nightly-2025-08-01 to prevent the upstream update breaking rustsbi. I would love to suggest that update the toolchain manually once or twice one mouth.

<details>

</details>

### DIFF
--- a/prototyper/prototyper/src/firmware/mod.rs
+++ b/prototyper/prototyper/src/firmware/mod.rs
@@ -28,7 +28,7 @@ pub struct BootHart {
 
 #[unsafe(naked)]
 #[unsafe(link_section = ".fdt")]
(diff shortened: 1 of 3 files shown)


luojia65 (approved):

guttatus (changes requested):
`pub extern "C" fn raw_fdt()` also needs to be changed

guttatus (approved):
LGTM!

PR #149, opened 2025-08-31
feat(prototyper): auto add reversed-memory device tree node

woshiluo:
As [docs](https://github.com/rustsbi/rustsbi/blob/main/prototyper/docs/booting-archlinux-in-qemu-using-edk2-and-rustsbi.md) said, prototyper will not add `reversed-memory` node automatically and may cause problem.

This pull request done the following jobs:

- Auto add `reversed-memory` node and protect sbi memory by it.

Wait for https://github.com/rustsbi/serde-device-tree/pull/17 for unit serialize support.


luojia65 (approved):

guttatus (approved):
LGTM

PR #154, opened 2025-09-08
fix(test-kernel):Update dependencies and Code clean

13584452567:


library/sbi-testing/src/spi.rs
```diff
@@ -35,7 +35,7 @@ pub fn test(hart_id: usize, mut f: impl FnMut(Case)) {
     fn ipi(hart_id: usize) -> ! {
         sbi::send_ipi(HartMask::from_mask_base(1 << hart_id, 0));
         // 必须立即触发中断，即使是一个指令的延迟，也会触发另一个异常
-        unsafe { core::arch::asm!("unimp", options(noreturn, nomem)) };
+        unsafe { core::arch::asm!(".word 0x00000000", options(noreturn, nomem)) };
```
luojia65, line 38: Explain why this change is necessary. The `unimp` instruction is deliberately designed to generate an illegal instruction in RISC-V assembly code.

prototyper/test-kernel/.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[build]
+target = "riscv64imac-unknown-none-elf"
```
luojia65, line 2: This file only works iff we run `cargo` command within the directory containing the `.cargo` path, in this case `prototyper/test-kernel`. Explain why we should have nested cargo config despite we have one in the project root.

luojia65 (changes requested):

PR #162, opened 2025-10-20
fix: hsm should return invalid when hartid is invalid

yizishun:
hsm call should not return success when the hart is disabled


woshiluo (approved):
LGTM.

luojia65 (approved):

PR #165, opened 2025-10-31
feat(prototyper): split boot hart to init hart and boot hart

woshiluo:
To make sbi can normal output when start with wrong params, like invalid `DynamicInfo`, this pull request split boot hart to init and boot hart.


luojia65:
Got correct output on QEMU Virt machine:
图片
The upstream code does not print any output, instead of the rich output for the firmware from this pull request.

luojia65 (approved):
Nice catch, LGTM

PR #169, opened 2025-11-16
fix(prototyper): remove a semicolon in `boot` function

crjg-k:
RustRover 2025.2.2 suggests that there is a type mismathcing: a redundant semicolon returns () type which does not match ! type stipulated in the signature of the function.

The original error messages are as follows:

```
Type mismatch [E0308]
Expected: !
Found: ()
```

image

### DIFF
--- a/prototyper/prototyper/src/sbi/trap/boot.rs
+++ b/prototyper/prototyper/src/sbi/trap/boot.rs
@@ -34,7 +34,7 @@ pub unsafe extern "C" fn boot() -> ! {
         "mret",
         locate_stack = sym trap_stack::locate,
         boot_handler = sym boot_handler,
-    );
+    )
 }
 
 /// Boot Handler.


Plucky923:
Thank you for the detailed explanation and your contribution. I have no objection to merging the code.
But just to clarify one point: the inspections and checks provided by RustRover are part of a proprietary JetBrains IDE (or plugin). We cannot reasonably require every contributor to install and use the same commercial IDE/plugin for code review or quality checks.
Therefore, when we need strict, enforceable code quality gates, the only practical and fair approach is to run them in CI/CD using fully open-source, headless, and freely available tools.
For that reason, I don’t think RustRover-specific inspections or its built-in test runner are suitable for integration into the official testing/CI pipeline of an open-source project.

zevorn (approved):
LGTM

luojia65 (approved):
We can head for generic proprietary CICD issues in the future.

Plucky923 (approved):

Placebo27 (approved):
There is no functional impact, but it does match the function signature.

PR #180, opened 2026-01-04
docs(prototyper): correct the comments of pmp setting

crjg-k:
According to the following code, the permission of segment `[sbi_start..sbi_rodata_start]` is `R`, while the another one `[sbi_rodata_end..sbi_end]` is same as above.

image


luojia65:
@woshiluo If this is the case we can accept this PR and then remove PMP comments in following contributions.

woshiluo (approved):

luojia65 (approved):

PR #197, opened 2026-03-18
fix(prototyper): make sbi_console_getchar non-blocking

### LINKED ISSUE #194: [BUG] prototyper: `sbi_console_getchar` blocks forever when no input available
## Summary

The legacy `sbi_console_getchar` (EID=0x2) implementation in RustSBI Prototyper blocks indefinitely when no console input is available, instead of returning -1 as specified by the SBI specification. This can cause a hart to hang permanently.

### 1. Clone and build RustSBI Prototyper

```bash
git clone https://github.com/rustsbi/rustsbi.git
cd rustsbi

# Install required Rust target
rustup target add riscv64imac-unknown-none-elf

# Build prototyper
cargo xtask prototyper
```

### 2. Create a minimal test payload

Create `test_getchar.S`:

```asm
.section .text
.globl _start
_start:
    # Call sbi_console_getchar (legacy extension)
    # EID = 0x2 (LEGACY_CONSOLE_GETCHAR)
    # According to SBI spec, this should return -1 if no input available
    li a7, 0x2
    li a6, 0x0
    ecall

    # If we reach here, getchar returned (correct behavior)
    # Print "OK" to indicate success
    li a7, 0x1          # legacy console putchar
    li a0, 0x4F         # 'O'
    ecall
    li a7, 0x1
    li a0, 0x4B         # 'K'
    ecall
    li a7, 0x1
    li a0, 0x0A         # newline
    ecall

    # Shutdown
    li a7, 0x08
    ecall

hang:
    j hang
```

Create `test_getchar.ld`:

```ld
OUTPUT_ARCH(riscv)
ENTRY(_start)

SECTIONS
{
    . = 0x80200000;
    .text : { *(.text) }
}
```

Compile:

```bash
riscv64-unknown-elf-gcc -nostdlib -nostartfiles -march=rv64imac -mabi=lp64 \
    -T test_getchar.ld -o test_getchar.elf test_getchar.S
riscv64-unknown-elf-objcopy -O binary test_getchar.elf test_getchar.bin
```

### 3. Run with QEMU

```bash
qemu-system-riscv64 \
    -machine virt \
    -nographic \
    -bios target/riscv64imac-unknown-none-elf/release/rustsbi-prototyper.bin \
    -kernel test_getchar.bin \
    -smp 1 \
    -m 256M
```

## Expected Behavior

According to the [SBI Specification](https://github.com/riscv-non-isa/riscv-sbi-doc), the legacy `sbi_console_getchar` function should:

- Return the character read from console (0-255) if available
- **Return -1 if no character is available** (non-blocking)

The test payload should:
1. Call `sbi_console_getchar`
2. Receive -1 (no input available)
3. Print "OK"
4. Shutdown

## Actual Behavior

The system hangs indefinitely at the `sbi_console_getchar` call. The "OK" message is never printed, and the system never shuts down.

You need to press `Ctrl+A X` to exit QEMU.

## Root Cause

In `prototyper/prototyper/src/sbi/console.rs`, lines 61-69:

```rust
pub fn getchar(&self) -> usize {
    let mut c = 0u8;
    let console = self.inner.lock();
    // Block until we successfully read 1 byte
    while console.read(core::slice::from_mut(&mut c)) != 1 {
        core::hint::spin_loop();
    }
    c as usize
}
```

The code explicitly blocks in a `while` loop until a character is read. This violates the SBI specification which requires non-blocking behavior.

## Suggested Fix

```rust
pub fn getchar(&self) -> usize {
    let mut c = 0u8;
    let console = self.inner.lock();
    if console.read(core::slice::from_mut(&mut c)) == 1 {
        c as usize
    } else {
        usize::MAX  // Return -1 to indicate no character available
    }
}
```

## Impact

- **Severity**: High
- **Type**: Denial of Service / Specification Violation

Any S-mode software (including operating systems) that calls `sbi_console_getchar` to check for available input will cause the hart to hang permanently. This is a common pattern in console drivers and bootloaders.

## Comparison with OpenSBI

OpenSBI correctly implements non-blocking behavior in `lib/sbi/sbi_console.c`:

```c
int sbi_getc(void)
{
    if (console_dev && console_dev->console_getc)
        return console_dev->console_getc();
    return -1;
}
```

And in `lib/utils/serial/uart8250.c`:

```c
static int uart8250_getc(void)
{
    if (get_reg(UART_LSR_OFFSET) & UART_LSR_DR)
        return get_reg(UART_RBR_OFFSET);
    return -1;  // Non-blocking: return -1 if no data
}
```

manchangfengxu:
## Summary
Fixes #194. Resolves the blocking behavior of `sbi_console_getchar` that violates the RISC-V SBI specification. Refactors `UartBflbWrap::read` to be a non-blocking implementation.

---

## What Changed

* **`SbiConsole::getchar`**: Updated the logic to check the read length `nread` from the underlying driver. If `nread` is 0 (no input), it immediately returns `usize::MAX` (representing `-1`) instead of spinning.

* **`UartBflbWrap::read`**: Refactored the driver logic by removing the `while` loop that waited for `receive_available_bytes()`. 
    * If `rx_available` is 0, it returns 0 immediately. 
    * Otherwise, it reads `min(rx_available, buf.len())` characters from the FIFO and returns the actual number of bytes read.

---

## Why

* Specification Compliance: The RISC-V SBI specification requires `sbi_console_getchar` to return `-1` if no character is available. The previous blocking implementation, included:UartBflbWrap::read , violated this requirement.
* Implementations such as OpenSBI's `uart8250`are non-blocking.
* Driver implementations, such as [RT-Thread's `bl_mcu_sdk`](https://github.com/RT-Thread/rt-thread/blob/master/bsp/bouffalo_lab/libraries/bl_mcu_sdk/drivers/lhal/src/bflb_uart.c#L236)
``` rust
ATTR_TCM_SECTION int bflb_uart_get(struct bflb_device_s *dev, uint8_t *data, uint32_t len)
{
    int ch = -1;
    uint32_t count = 0;

    while (count < len) {
        if ((ch = bflb_uart_getchar(dev)) < 0) {
            break;
        }
        data[count] = ch;
        count++;
    }
    return count;
}
```
* The [BouffaloSDK UART - poll example](https://bl-mcu-sdk.readthedocs.io/zh-cn/latest/samples/peripherals/uart/uart_poll.html) 
``` rust
int ch;
while (1) {
    ch = bflb_uart_getchar(uartx);
    if (ch != -1) {
        bflb_uart_putchar(uartx, ch);
    }
}
```

---

**Fixes: #194**

### DIFF
--- a/prototyper/prototyper/src/platform/console.rs
+++ b/prototyper/prototyper/src/platform/console.rs
@@ -106,13 +106,11 @@ impl UartBflbWrap {
 impl ConsoleDevice for UartBflbWrap {
     fn read(&self, buf: &mut [u8]) -> usize {
         let uart = unsafe { &(*self.inner) };
-        while uart.fifo_config_1.read().receive_available_bytes() == 0 {
-            core::hint::spin_loop();
+        let rx_available = uart.fifo_config_1.read().receive_available_bytes() as usize;
+        if rx_available == 0 {
+            return 0;
         }
-        let len = core::cmp::min(
-            uart.fifo_config_1.read().receive_available_bytes() as usize,
-            buf.len(),
-        );
+        let len = core::cmp::min(rx_available, buf.len());
         buf.iter_mut()
             .take(len)
             .for_each(|slot| *slot = uart.fifo_read.read());
--- a/prototyper/prototyper/src/sbi/console.rs
+++ b/prototyper/prototyper/src/sbi/console.rs
@@ -77,17 +77,13 @@ impl SbiConsole {
 
     /// Reads a single character from the console.
     ///
-    /// This method will block until a character is available to be read.
-    ///
     /// # Returns
-    /// The read character as a usize
+    /// Returns the character as a usize on success, or '-1' for failure.
     #[inline]
     pub fn getchar(&self) -> usize {
         let mut c = 0u8;
-        while self.inner.lock().read(core::slice::from_mut(&mut c)) == 0 {
-            core::hint::spin_loop();
-        }
-        c as usize
+        let nread = self.inner.lock().read(core::slice::from_mut(&mut c));
+        if nread == 1 { c as usize } else { usize::MAX }
     }
 
     // Rejects buffers that this firmware cannot safely turn into raw slices.


woshiluo (approved):
Thank you

luojia65 (approved):
LGTM

PR #213, opened 2026-05-25
chore(docker): add development image

Polardist:
We can use docker to develop rustsbi now :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,11 +25,13 @@ The RustSBI Prototyper and ArceBoot requires nightly Rust version of `rustc 1.97
 
 ## Build this project
 
-### For firmware user
+### Native development
+
+#### For firmware user
 
 If you need to build RustSBI firmware for M-mode, please refer to the [RustSBI Prototyper](firmware/README.md) documentation.
 
-### For library user
+#### For library user
 
 RustSBI is usually used as a library or dependency. If you wish to, you may build RustSBI library itself using the
 following command:
@@ -52,6 +54,26 @@ cargo build --target riscv64imac-unknown-none-elf
 
 The target platform of RISC-V platform firmware is usually a bare metal target.
 Under normal circumstances these targets in Rust would start with `riscv??-` and end with `-none-elf`.
+
+### Or use the Docker development environment
+
+The repository provides a Docker image for development on hosts that do not match the CI environment.
+It includes the pinned nightly toolchain, RISC-V targets, QEMU, U-Boot tools, `cargo-binutils` and `axconfig-gen`.
+
+For a one-off check, run:
+
+```bash
+docker build -t rustsbi-dev .
+docker run --rm -v "$PWD:/workspace" -w /workspace rustsbi-dev cargo check
+```
+
+For an interactive shell, run:
+
+```bash
+docker run --rm -it -v "$PWD:/workspace" -w /workspace rustsbi-dev bash
+```
+
+The source tree is mounted at `/workspace`; changes and build artifacts stay on the host.
 
 ## Features
 
--- a/README.md
+++ b/README.md
@@ -25,6 +25,26 @@ The RustSBI Prototyper and ArceBoot requires nightly Rust version of `rustc 1.97
 
 ## Build this project
 
+### With Docker
+
+The repository provides a Docker image for development on hosts that do not match the CI environment.
+It includes the pinned nightly toolchain, RISC-V targets, QEMU, `cargo-binutils` and `axconfig-gen`.
+
+For a one-off check, run:
+
+```bash
+docker build -t rustsbi-dev .
+docker run --rm -v "$PWD:/workspace" -w /workspace rustsbi-dev cargo check
+```
+
+For an interactive shell, run:
+
+```bash
+docker run --rm -it -v "$PWD:/workspace" -w /workspace rustsbi-dev bash
+```
+
+The source tree is mounted at `/workspace`; changes and build artifacts stay on the host.
+
 ### For firmware user
 
 If you need to build RustSBI firmware for M-mode, please refer to the [RustSBI Prototyper](prototyper/README.md) documentation.
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,14 @@
+/target
+/.git
+/.idea
+/.DS_Store
+/.codex
+/.humanize
+arceboot/mnt/
+arceboot/myramdisk/
+arceboot/mnt_fat32/
+arceboot/edk2/
+arceboot/*.img
+arceboot/*.cpio
+qemu-logs/
+/.openeuler-aia/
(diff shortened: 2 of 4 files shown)


woshiluo:
Since there is also a `rust-toolchain.toml` file in project's root directory, do we still need to pin rust version manually in Dockerfile?

Polardist:
```bash
root@83444c84d758:/workspace#  rustup target list --installed
aarch64-unknown-linux-gnu
riscv64gc-unknown-none-elf
riscv64imac-unknown-none-elf
```

Now we good

woshiluo (approved):
LGTM

README.md
```diff
@@ -25,6 +25,26 @@ The RustSBI Prototyper and ArceBoot requires nightly Rust version of `rustc 1.97
 
 ## Build this project
 
+### With Docker
```
luojia65: This implies we provide Docker as the primary development method, rather than a clean-room development with only the Rust toolchain installed. We should first provide cargo and rustc only development, then we provide Docker.

Some of our downstream developers do not specialize in computer technology (but instead in aviation, medical, or automation etc.), requiring the use of Docker will eventually lead to a software that is hard to begin with; compiling and using _firmware_ instead of _user-mode software_ is already a bar of SBI firmware development, we shouldn't set it higher.

Proposed change:
```
## Build this project
### For firmware users (using `cargo` only)
### For library users (using `cargo` only)
### Using Docker (alternative approach)
```

luojia65 (changes requested):
Code looks fine to me, there should be a change to README.

Polardist:
You're right, let me close this PR then. That way, the downstream won't have any psychological burden.

Polardist:
CC @Plucky923

Plucky923:
@Polardist Thank you for your contribution!

@luojia65 Regarding the development workflow, I have a few thoughts.

> This implies we provide Docker as the primary development method, rather than a clean-room development with only the Rust toolchain installed.

I don't think providing Docker here necessarily implies that Docker becomes the only or mandatory development method.

The native Cargo workflow can remain fully supported, while Docker provides a reproducible development environment alongside it. These two approaches are not mutually exclusive.

In fact, containerized development environments are already commonly used in low-level systems projects. For example, Firecracker uses a development container for building and testing, seL4 recommends Docker for isolating build dependencies, and Asterinas provides a Docker-based kernel development environment.

The main purpose of Docker here is not to require downstream developers to learn an additional tool, but to package the required Rust toolchain, RISC-V targets, QEMU, and other dependencies into a reproducible environment. Developers who prefer a native Rust toolchain can still use the native workflow directly.

I also think there are two separate dimensions here:

* firmware/library users describe **what** is being built;
* native/Docker environments describe **how** the development environment is provided.

So rather than deciding whether Docker is the “primary” or an “alternative” approach, perhaps the README could simply document both development workflows, for example:

```text
## Build this project

### Native development
#### Firmware
#### Library

### Docker development environment
```

This keeps the native workflow straightforward while also providing a convenient and reproducible environment for developers who want it.

Plucky923 (commented):
@Polardist Thanks for adding the development image. I think it would be useful to add CI coverage for the Docker workflow as part of this PR.

At the moment, the README mainly verifies the container with `cargo check`. That is useful as a basic sanity check, but I don't think it is sufficient to verify a RustSBI development environment. In particular, `cargo check` does not exercise the complete firmware toolchain, such as cross-target compilation, linker scripts, binary generation, `cargo-binutils`, or QEMU boot.

Since this image is intended to provide a supported development environment, I think the CI should verify the following levels:

1. **The Docker image can be built successfully.**

   ```bash
   docker build -t rustsbi-dev .
   ```

   This verifies that the system dependencies, Rust toolchain, RISC-V targets, and Cargo tools can still be installed from a clean environment.

2. **The main tools required by the development environment are actually available inside the image.**

   For example:

   ```bash
   docker run --rm rustsbi-dev bash -lc '
       rustc --version &&
       cargo --version &&
       qemu-system-riscv64 --version &&
       rustup target list --installed &&
       cargo objcopy --version &&
       axconfig-gen --version
   '
   ```

   This is mostly a diagnostic/sanity check, but it makes failures easier to understand when the image becomes stale.

3. **RustSBI firmware should actually be built inside the container.**

   We already have native CI coverage for the Prototyper firmware modes, so I think the Docker CI should reuse the same workflow rather than only running `cargo check`.

   For example, for the current Prototyper interface:

   ```bash
   cargo test-kernel

   # Dynamic firmware
   cargo prototyper

   # Jump firmware
   cargo prototyper --jump

   # Payload firmware
   cargo prototyper \
       --payload \
       target/riscv64imac-unknown-none-elf/release/rustsbi-test-kernel.bin
   ```

   This verifies considerably more than `cargo check`: the RISC-V targets, build scripts, linker scripts, ELF/bin generation, `cargo-binutils`, and the firmware build pipeline are all exercised.

4. **At least one firmware image should actually boot in QEMU.**

   I don't think we need to duplicate the entire existing QEMU boot matrix inside Docker. The native CI already tests the different firmware modes and both test/bench kernels.

   For the Docker workflow, one representative end-to-end smoke test should be enough to answer the Docker-specific question:

   > Can a developer with only Docker installed build the RustSBI development image, build a real RustSBI firmware image inside it, and successfully boot that firmware in QEMU?

   Since the repository already provides the QEMU boot script, we could simply reuse it, for example:

   ```bash
   .github/scripts/prototyper-qemu-boot.sh payload test
   ```

   So the complete Docker smoke test could look roughly like:

   ```yaml
   - name: Build development image
     run: docker build -t rustsbi-dev .

   - name: Build and boot RustSBI firmware in Docker
     run: |
       docker run --rm \
         -v "$PWD:/workspace" \
         -w /workspace \
         rustsbi-dev \
         bash -lc '
           set -euxo pipefail

           cargo test-kernel

           cargo prototyper
           cargo prototyper --jump
           cargo prototyper \
             --payload \
             target/riscv64imac-unknown-none-elf/release/rustsbi-test-kernel.bin

           .github/scripts/prototyper-qemu-boot.sh payload test
         '
   ```

I think this gives us a useful separation of responsibilities:

* the existing native CI verifies the full RustSBI firmware/QEMU test matrix;
* the Docker CI verifies that the container itself remains a complete and usable RustSBI development environment.

This is also important for long-term maintenance. A Dockerfile may continue to build successfully even after the actual firmware development workflow has gained a new host dependency or tool requirement. An end-to-end firmware smoke test would catch that kind of environment drift automatically.

So I would suggest adding at least:

```text
Docker build
    ↓
build representative RustSBI firmware
    ↓
boot one firmware configuration in QEMU
    ↓
verify the test kernel completes successfully
```

rather than treating `cargo check` alone as validation of the development image.

luojia65:
@Polardist : Please follow instructions from @Plucky923

Polardist:
Now we good. CC @Plucky923 , @luojia65

Plucky923 (approved):
LGTM!

PR #215, opened 2026-07-28
Refactor prototyper

Plucky923:
This PR is intended solely as a proof-of-concept (PoC) and reference implementation. The implementation in this PR was developed entirely by AI and is not intended to be merged as-is.

The follow-up work #219 will be tracked and implemented through separate issues and PRs. Once those issues are resolved and the corresponding changes have landed, this PR will be closed.

@luojia65 @zevorn Thanks to everyone who reviewed this PR and provided valuable feedback and suggestions.


library/sbi-testing/src/protocol.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// Stable matrix shard identifier.
+    pub shard: &'a str,
+    /// Stable invocation identifier within the shard.
+    pub run: &'a str,
+    /// One-based execution attempt. Acceptance runs always use one.
+    pub attempt: u32,
+    /// Deterministic schedule or input seed.
+    pub seed: u64,
+    /// SHA-256 digest of the ordered expected case identifiers.
+    pub digest: &'a str,
```
luojia65, line 20: Are those two variables necessary in sbi-testing framework?

prototyper/devkit/cargo-prototyper/src/artifact.rs
```diff
@@ -0,0 +1,186 @@
...
+        .map_err(|error| Error::io("start process", error))?;
+    if status.success() {
+        Ok(())
+    } else {
+        Err(Error::ProcessFailed {
+            program: program.into(),
+            code: status.code(),
+        })
+    }
+}
```
luojia65, line 186: If `rust-objcopy` or `mkimage` is not installed, this function only returns error at "start process" with "No such file or directory", even without printing which binary is not installed, instead of providing helpful instructions:

```
$ cargo prototyper run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/cargo-prototyper run`
cargo prototyper: building test for riscv64imac-unknown-none-elf
    Finished `release` profile [optimized + debuginfo] target(s) in 0.08s
cargo prototyper: start process: No such file or directory (os error 2)
```

We should provide users with helpful messages here, if we detect that such a program does not exist (not installed). We can print help message for installing `rust-objcopy` (hint the user to install `cargo install cargo-binutils`), or `mkimage` (install package `u-boot-tools` from Linux distrbutions, or build from scratch on Windows platform).

For example:

```
cargo prototyper: `rust-objcopy` cannot be found; raw system error: No such file or directory (os error 2)
cargo prototyper: help: if not installed, install one with `cargo install cargo-binutils`.
```

prototyper/devkit/cargo-prototyper/src/runner.rs
```diff
@@ -0,0 +1,305 @@
...
+
+/// Runs a resolved launch plan without inspecting build metadata.
+pub fn run(plan: &LaunchPlan) -> Result<Outcome, Error> {
+    let mut child = Command::new(&plan.program)
+        .current_dir(&plan.working_directory)
+        .args(&plan.arguments)
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()
+        .map_err(|error| Error::io("start runner", error))?;
```
luojia65, line 159: We'd provide with more friendly help messages.

```
$ cargo prototyper run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/cargo-prototyper run`
cargo prototyper: building test for riscv64imac-unknown-none-elf
    Finished `release` profile [optimized + debuginfo] target(s) in 0.08s
cargo prototyper: binary /data/luojia/RustSBI/rustsbi/target/riscv64imac-unknown-none-elf/release/rustsbi-test-kernel.bin
cargo prototyper: building firmware for riscv64imac-unknown-none-elf
    Finished `release` profile [optimized + debuginfo] target(s) in 0.09s
cargo prototyper: binary /data/luojia/RustSBI/rustsbi/target/riscv64imac-unknown-none-elf/release/rustsbi-firmware-fw-payload.bin
cargo prototyper: start runner: No such file or directory (os error 2)
```

Instead of:

```
cargo prototyper: start runner: No such file or directory (os error 2)
```

We can provide with:

```
cargo prototyper: `qemu-system-riscv64` cannot be found: No such file or directory (os error 2)
cargo prototyper: help: if not installed, install one using `apt install qemu-system-riscv64`
```

The help messages can be summerized in one Rust module, then we can provide detailed help messages according to Linux distributions, macOS versions or Windows environments.

prototyper/devkit/cargo-prototyper/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "cargo-prototyper"
+version = "0.1.0"
```
luojia65, line 3:
```suggestion
version = "0.0.0"
```

Defaults to `0.0.0` for all packages if `publish = false`.

prototyper/devkit/qemu-virt/firmware.ld
luojia65, line 1: This file can be moved back to `prototyper/firmware` either as a separate linker script, or a built-in linker inlined in `build.rs` file. Or for dynamic firmware (on real hardware, not on fixed QEMU platform) this file is still inside a folder called `qemu-virt` which may be misleading to new developers.

prototyper/devkit/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+普通 `cargo check` 仍可用于快速检查 Rust 源码；只有
+`cargo prototyper build` 承诺生成采用受审链接布局、完成 ELF 到裸二进制转换并
+使用稳定名称的可启动最终镜像。
+
+不带额外参数的 `run` 会自动构建默认 S-mode 测试 payload、嵌入固件并启动
+QEMU。只有运行自定义 payload 时才需要传 `--payload <BIN>`。
+
+旧的 `cargo xtask prototyper`、`cargo xtask test` 和
+`cargo xtask bench` 暂时保留，但只会显示弃用提示并转交给这个工具。
```
luojia65, line 30: Consider either rename `cargo prototyper` into `cargo xtask` and remove original `cargo xtask` in the next pull request?

prototyper/docs/booting-ubuntu-24.04.3-in-qemu-using-edk2-and-rustsbi.md
```diff
@@ -30,16 +36,18 @@ xz -d ubuntu-24.04.3-preinstalled-server-riscv64.img.xz
...
 
 ```bash
-git clone https://github.com/rustsbi/rustsbi.git --depth 1
-cd rustsbi
-cargo prototyper
+git clone https://github.com/rustsbi/rustsbi.git --depth 1 "$WORKSHOP_DIR/rustsbi"
+(
+    cd "$WORKSHOP_DIR/rustsbi"
+    cargo prototyper
+)
```
luojia65, line 43: Is the `(` and `)` necessary?

prototyper/docs/firmware-internal-privilege-separation.md
```diff
@@ -0,0 +1,45 @@
...
+RustSBI Prototyper 将这种问题称为**固件内特权分离架构**。这里的“固件内”很
+重要：分离发生在同一份 M 态固件内部，而不是以另一种方式重复 RISC-V 的 M/S
+硬件特权级。`machine` 层持有直接作用于机器状态的机制权力；上层
+`rustsbi-prototyper` 则以安全 Rust 编写平台策略和 SBI 服务。上层能够表达“启动
+一个 hart”“设置定时器”或“处理一项 SBI 调用”等意图，但不能得到裸 CSR、物理
+指针、MMIO 寄存器窗口之外的地址，或决定如何恢复一份 trap frame。
+
+这不是把上层 M 态代码变成不可信代码的硬件安全沙箱。两层仍共享同一个特权级、
+地址空间和可信计算基；带有任意 `unsafe` 的恶意 M 态代码依然可以越过这些
+约束。该架构提供的是更窄的**软件权威边界**：把高风险机制集中到较小、可审计的
```
luojia65, line 29:
```suggestion
约束。该架构提供的是更窄的**软件权限边界**：把高风险机制集中到较小、可审计的
```

prototyper/drivers/sifive-test/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "rustsbi-prototyper-sifive-test"
+version = "0.0.0"
```
luojia65, line 3: For packages that does not set `publish = false`, add descriptions.

```toml
description = "SiFive Test device driver for RustSBI Prototyper"
```

Add authors, list yourself as an author.

```toml
authors = ["Your Name <your_email@example.com>"]
```

prototyper/drivers/uart-bl808/src/lib.rs
luojia65, line 1: I understand the demand of formal verification when we independently rewrite drivers for RustSBI Prototyper instead of reusing community `bouffalo-hal`. In the future if we contibute to the community to have a paritally formal verified `bouffalo-hal` (in this case, at least UART peripheral is verified), this package would be used as an adapter for existing packages which avoids extra code or proof in the view of whole Rust community.

prototyper/drivers/uart-pl011/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+    const CONTROL: usize = 0x30;
+    const INTERRUPT_MASK: usize = 0x38;
+    const INTERRUPT_CLEAR: usize = 0x44;
+    const RECEIVE_EMPTY: u32 = 1 << 4;
+    const TRANSMIT_FULL: u32 = 1 << 5;
+}
+
+/// Initializes and binds one already-owned PL011 register window.
+pub fn bind(io: IoMem) -> Result<Console, IoMemError> {
+    io.validate::<u32>(Register::INTERRUPT_CLEAR)?;
```
luojia65, line 28: Should we validate all the registers, or a range of registers, instead of one register? It costs the same time (the same O(1) bound checks in `IoMem::address`) but looks more logical to check all I/O registers are accessible before read or write to them.

prototyper/firmware/src/device_tree/aia.rs
luojia65, line 1: Does the QEMU `virt` machine include AIA peripheral? We may note that this module is not checked. Or in following pull requests we introduce gevico/qemu or upstream QEMU to test AIA environment.

prototyper/firmware/src/sbi/pmu/mod.rs
```diff
@@ -0,0 +1,582 @@
...
+            Err(error) => return error.into(),
+        };
+        let mut state = match self.current_state() {
+            Ok(state) => state,
+            Err(error) => return error.into(),
+        };
+
+        if flags.contains(flags::CounterCfgFlags::SKIP_MATCH) {
+            let Some(index) = selection.first() else {
+                machine::abort(|| {});
```
luojia65, line 277: There are multiple `machine::abort(|| {})` in this module. Describe why machine aborts here, instead of returning any SbiRet error values.

prototyper/firmware/src/sbi/pmu/selection.rs
```diff
@@ -0,0 +1,54 @@
...
+use sbi_spec::binary::Error;
+
+#[derive(Clone, Copy)]
+pub(super) struct CounterSelection {
+    base: usize,
+    mask: usize,
+}
+
+impl CounterSelection {
+    pub(super) fn new(base: usize, mask: usize, total: usize) -> Result<Self, Error> {
```
luojia65, line 14: Leave documents here describing this `CounterSelection` only selects counters described with `base` and `mask` within counter index range of `[0, total)`.

prototyper/firmware/src/sbi/pmu/state.rs
```diff
@@ -0,0 +1,65 @@
...
+        counters: &PerformanceCounters,
+    ) -> Result<(), CounterError> {
+        if self.initialized {
+            return Ok(());
+        }
+        for index in 0..counters.count() {
+            let info = counters.info(index)?;
+            let event_index = match info.csr_number() {
+                0x0c00 => Some(hardware_event::CPU_CYCLES),
+                0x0c02 => Some(hardware_event::INSTRUCTIONS),
```
luojia65, line 47: Are those hex digits `cycle` and `instret`? Define them using `const CSR_CYCLE: u16` etc.

prototyper/firmware/src/sbi/console.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+}
+
+impl rustsbi::Console for DebugConsole {
+    fn write(&self, bytes: Physical<&[u8]>) -> SbiRet {
+        let mut source = match self.memory.reader(
+            bytes.phys_addr_lo(),
+            bytes.phys_addr_hi(),
+            bytes.num_bytes(),
+        ) {
```
luojia65:
```suggestion
        let mut source = match self.memory.reader(bytes) {
```

Rearrange `fn reader` and `fn writer` to accept `Physical<T>`.

prototyper/firmware/src/sbi/ipi.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+pub(super) fn targets(mask: HartMask) -> HartTargets {
+    let (bits, base) = mask.into_inner();
+    if base == usize::MAX {
+        HartTargets::all_available()
+    } else {
+        HartTargets::selected(bits, base)
+    }
+}
```
luojia65: We may found that `HartMask` and `HartTargets` provides the same kind of information, thus may be redundant. Can we just remove `HartTargets`, `fn targets` and use `mask` instead of `target(mask)` in our code? We can keep argument name `target: HartMask` in those functions.

prototyper/firmware/src/sbi/timer.rs
```diff
@@ -0,0 +1,19 @@
...
+    pub(super) fn new(timer: MachineTimer) -> Self {
+        Self { timer }
+    }
+}
+
+impl rustsbi::Timer for Timer {
+    fn set_timer(&self, deadline: u64) {
+        self.timer.set_deadline(deadline);
+    }
+}
```
luojia65, line 19: Is the reason we don't `impl rustsbi::Timer for MachineTimer` in crate `machine` here for the internal interface of firmware to fulfill design goal of formal verification?

prototyper/firmware/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    let interrupts = match device_tree::install_interrupts(&mut boot, &mut tree, &discovered_harts)
+    {
+        Ok(interrupts) => interrupts,
+        Err(_) => machine::abort(|| {}),
+    };
+    let console = match device_tree::install_console(&mut boot, &mut tree) {
+        Ok(console) => console,
+        Err(_) => machine::abort(|| {}),
+    };
```
luojia65, line 44: Can we install console first, then we can print error message if any following initialization stages returns `Err(e)`? In such procedure we won't abort without any messages if any of those stages fail.

prototyper/firmware/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+uart-bl808 = { package = "rustsbi-prototyper-uart-bl808", path = "../drivers/uart-bl808" }
+sifive-test = { package = "rustsbi-prototyper-sifive-test", path = "../drivers/sifive-test" }
+
+[[bin]]
+name = "rustsbi-firmware"
+test = false
+bench = false
+
+[features]
+default = []
```
luojia65, line 31: If we support RVA23 by default, is it possible to enable hypervisor by default?

```suggestion
default = ["hypervisor"]
```

prototyper/machine/macros/Cargo.toml
luojia65, line 1: Can we name this package `rustsbi-prototyper-machine-macros` in the path `prototyper/machine/macros`?

prototyper/machine/src/hart/admission/mod.rs
```diff
@@ -0,0 +1,290 @@
...
+use super::fence::RemoteFenceRequest;
+use crate::hart::{HartStatus, HartTargets};
+
+#[cfg(test)]
+mod audit;
+mod lifecycle;
+mod remote_fence;
+
+#[derive(Clone, Copy, Debug, Default, Eq, Hash, PartialEq)]
+pub(crate) struct HartSet(pub(super) u128);
```
luojia65: Reuse `sbi_spec::HartMask`? This sturcture is maybe redundant, and it only supports up to 128 harts.

prototyper/machine/src/hart/control.rs
```diff
@@ -0,0 +1,207 @@
...
+use alloc::sync::Arc;
+
+use super::admission::{AdmissionError, ClaimedWork};
+use super::instructions::{clear_supervisor_ipi, current_hart_id, wait_for_wake_event};
+use super::protocol::{HartAdmission, map_hart_error};
+use crate::boot::NextStage;
+
+/// Ratified hart lifecycle state.
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub enum HartStatus {
```
luojia65: Move to `sbi-spec` as `sbi_spec::hsm::HartState`?

prototyper/machine/src/interrupt/aia/aplic.rs
luojia65, line 1: Reuse `riscv-aia` crate in `library/riscv-aia` for most of aplic and imsic structures, constants and enums.

prototyper/machine/src/pmu/access.rs
```diff
@@ -0,0 +1,68 @@
...
+/// The counter masks come only from the current hart's successfully probed
+/// Zicntr/Zihpm state. Time remains readable in S-mode so the trap path can
+/// emulate unavailable direct reads without granting arbitrary CSR access.
+pub(super) fn prepare_counter_access(
+    mode: NextMode,
+    counters: &PerformanceCounters,
+) -> Result<(), CounterError> {
+    let accessible = counters.accessible_mask()? as usize;
+    let (machine_access, supervisor_access) = access_masks(mode, accessible);
+    let actual_machine = write_checked::<MCOUNTEREN>(machine_access, mode)?;
```
luojia65, line 23: Reuse `riscv::register::mcounteren`: https://docs.rs/riscv/latest/riscv/register/mcounteren/index.html.

prototyper/machine/src/power/terminal.rs
```diff
@@ -0,0 +1,150 @@
...
+    CONTROL.get().map(Box::as_ref)
+}
+
+pub(crate) fn begin_power_transition() -> bool {
+    STATE.begin_power()
+}
+
+pub(crate) fn is_terminal() -> bool {
+    STATE.terminal()
+}
```
luojia65, line 91: Document those frequent-used internal functions.

prototyper/machine/src/trap/frame.rs
```diff
@@ -0,0 +1,505 @@
...
+                x24: 0,
+                x25: 0,
+                x26: 0,
+                x27: 0,
+                x28: 0,
+                x29: 0,
+                x30: 0,
+                x31: 0,
+            },
+            mepc,
```
luojia65, line 280: Add `const fn zeroed() -> Self` for `Registers`.

prototyper/machine/src/memory.rs
```diff
@@ -0,0 +1,297 @@
...
+        Ok(Self { ranges })
+    }
+
+    /// Validates a complete physical range for machine reads.
+    pub fn reader(
+        &self,
+        base_addr_lo: usize,
+        base_addr_hi: usize,
+        requested_len: usize,
+    ) -> Result<Reader<'_>, MemoryError> {
```
luojia65: Use `sbi_spec::Physical` for `reader` and `writer`.

```suggestion
    /// Validates a complete physical range for machine reads.
    pub fn reader<T>(&self, bytes: Physical<T>) -> Result<Reader<'_>, MemoryError> {
```

luojia65 (changes requested):
Awesome work! I'd like to provide with multiple suggestions here to make it better.

prototyper/devkit/qemu-virt/firmware.ld
Plucky923: > This file can be moved back to `prototyper/firmware` either as a separate linker script, or a built-in linker inlined in `build.rs` file. Or for dynamic firmware (on real hardware, not on fixed QEMU platform) this file is still inside a folder called `qemu-virt` which may be misleading to new developers.

 Thanks for the suggestion. This design moved the linker script out of prototyper/firmware because the image layout is selected by cargo-prototyper together with the M-test, test-kernel, and benchmark linker scripts. I therefore consider it a devkit platform asset rather than firmware policy or source code (I think the firmware folder should only contain source code related to policy; the linker script is clearly needed for building the binary).
 
Inlining this relatively long script in build.rs mixes Rust build logic with linker-layout definitions and makes both harder to read and review.

The ownership could be made clearer for future platform support. Moving the linker-script group under prototyper/devkit/platforms/qemu-virt/ may express this intent better. When real-hardware support is introduced, we can add a platform-specific linker script or extract a genuinely shared layout then.

But as for which specific directory to place it in, I believe it needs to be separated from the firmware and then uniformly selected by the build tool. Regarding the developer issue, are there any other solutions?

prototyper/devkit/README.md
Plucky923: The original xtask should eventually be removed, but its current command model first needs to be migrated rather than directly renamed.

On the main branch, the xtask mixes different command categories: prototyper and arceboot identify products, while test and bench identify image roles. In addition, cargo xtask test only builds, converts, and optionally packages rustsbi-test-kernel; the actual QEMU test execution is handled separately by CI scripts.

The existing Cargo aliases already expose these workflows as cargo prototyper, cargo test-kernel, and cargo bench-kernel. This indicates that the generic xtask interface does not clearly describe the affected component or the performed operation.

A staged migration would avoid breaking existing workflows. The current pull request can retain the xtask commands as compatibility forwarders. A follow-up pull request can introduce product-scoped commands, migrate the remaining ArceBoot workflow, update CI and documentation, and then remove the original xtask and obsolete aliases.

Simply renaming cargo prototyper to cargo xtask would preserve the existing command ambiguity rather than resolve it.

prototyper/drivers/uart-pl011/src/lib.rs
Plucky923: The current code only checks the last register. This is enough for the current layout, but it is not clear and could be easy to get wrong when the driver changes. The code will be changed to check the whole PL011 register range before initializing the UART.

prototyper/firmware/src/sbi/pmu/mod.rs
Plucky923: These aborts are intentional for soundness. A normal PMU failure returns an SbiRet error if the previous state can be restored. If rollback also fails, the firmware can no longer guarantee that its PMU state matches the hardware, so continuing execution would be unsafe. The remaining aborts indicate internal states that should be impossible after validation. Comments will be added to explain these cases.

prototyper/firmware/src/sbi/pmu/state.rs
Plucky923: These are the cycle and instret CSR numbers. A dedicated CSR-number type with named constants would improve readability and prevent accidental mixing with other integer identifiers. This will be reworked in a follow-up revision.

library/riscv-aia/src/lib.rs
```diff
@@ -167,32 +172,13 @@ mod tests {
...
         let iid = Some(Iid::MSOFT);
-        match iid {
-            Some(Iid::MTIMER) => unreachable!(),
-            Some(Iid::MSOFT) => assert!(true),
-            Some(Iid::MEXT) => unreachable!(),
-            // Redundant, can be `_ => unreachable!()` in real use.
-            Some(_) | None => unreachable!(),
-        }
+        assert!(matches!(iid, Some(Iid::MSOFT)));
     }
```
luojia65, line 182: Those functions illustrates how `Iid` would be used in real code, and ensure such if-else and match functions works. We can add this as a comment so that LLM would not try to simplify code like this.

luojia65 (changes requested):

prototyper/machine/src/entry/dynamic.rs
```diff
@@ -0,0 +1,95 @@
...
+use crate::config::BOOT_STACK_SIZE;
+
+use super::stacks::{BOOT_STACK, HART_STACKS, HartStack};
+use super::{
+    EARLY_FAILED, EARLY_INITIALIZING, EARLY_READY, EARLY_STATE, RUNTIME_READY, RUNTIME_STATE,
+};
+use super::{allocator, enter_policy, fail_stop, image, relocation, warm_entry};
+
+const MSTATUS_MPRV: usize = 1 << 17;
+const DYNAMIC_WORD_COUNT: usize = 6;
```
zevorn, line 14: [P1] Preserve FW_DYNAMIC v1's five-word ABI

`DYNAMIC_WORD_COUNT = 6` makes `dynamic.S` bounds-check and load word 5 before it inspects `version`. OpenSBI v0.5's v1 structure has exactly five words; `boot_hart` was added only for v2. A valid v1 block at the end of a mapped page can therefore fault before this firmware installs a trap vector, and secondary harts compare unrelated memory. Read/snapshot the five common words first, then access `boot_hart` only for v2 (using `-1` for v1), and add a page-boundary v1 test. See the [v1 layout](https://github.com/riscv-software-src/opensbi/blob/v0.5/include/sbi/fw_dynamic.h).

prototyper/machine/src/hart/warm.rs
```diff
@@ -0,0 +1,81 @@
...
+    if runtime.prepare_current_hart().is_err() {
+        crate::trap::abort();
+    }
+    enable_notification(runtime.notification());
+
+    loop {
+        if crate::power::is_terminal() {
+            crate::power::halt();
+        }
+        let Some(mode) = runtime.pending_start_mode(hart_id) else {
```
zevorn, line 26: [P1] Close the check-before-WFI lost-wakeup window

Notifications are globally enabled here. If `start()` publishes and rings after `pending_start_mode()` returns `None` but before the following `wfi`, the trap handler can claim/clear that notification and return to `wfi`; the start request is not part of the ordinary IPI drain, so the hart may then sleep forever while the source hart waits for preparation. Keep global MIE clear across the state check and use a pending-event-safe wait/claim/recheck protocol, with a test for this interleaving.

prototyper/machine/src/pmp/state.rs
```diff
@@ -0,0 +1,179 @@
...
+    #[doc(hidden)]
+    pub fn empty() -> Self {
+        Self { grants: Vec::new() }
+    }
+
+    /// Adds one exact lower-privilege grant.
+    #[doc(hidden)]
+    pub fn grant(&mut self, range: Range<usize>, permissions: Permissions) -> Result<(), PmpError> {
+        let region = Region::new(range.start, range.end)?;
+        if permissions.is_empty()
```
zevorn, line 42: [P1] Reject unknown permission bits before encoding pmpcfg

`Permissions` and `grant` are public, and safe code can construct `Permissions::from_bits_retain(0x80)`. This check accepts it, after which `Entry::config_byte()` ORs bit 7 directly into `pmpcfg`, locking that PMP entry until reset; A/reserved bits can be injected the same way. Reject `permissions.bits() & !Permissions::all().bits() != 0` at this boundary and test every unknown bit before retaining the value.

prototyper/machine/src/hart/local.rs
```diff
@@ -0,0 +1,415 @@
...
+        if !slot.try_claim() {
+            interrupts.restore();
+            return Err(HartLocalError::Borrowed);
+        }
+        Ok(HartLocalGuard { slot, interrupts })
+    }
+}
+
+/// One lifetime-bound exclusive borrow of current-hart subsystem state.
+pub struct HartLocalGuard<'local, T> {
```
zevorn, line 276: [P2] Keep this CSR-restoring guard on its originating hart

For `T: Send`, this guard is automatically `Send` because `LocalSlot<T>: Sync` and `LocalInterrupts` contains only `usize`. Safe scoped code can move it to another hart; `Drop` then releases the source slot but restores the source hart's saved `mie` value on the destination, leaving both harts with incorrect interrupt state. Add a non-`Send` marker to the guard/token, document same-hart drop, and add a compile-time auto-trait assertion.

prototyper/machine/src/interrupt/clint.rs
```diff
@@ -0,0 +1,330 @@
...
+    };
+    let Some(offset) = clint.mtimecmp(hart) else {
+        return;
+    };
+    let _ = clint.registers.write_once(offset, u32::MAX);
+    let _ = clint
+        .registers
+        .write_once(offset + 4, (deadline >> 32) as u32);
+    let _ = clint.registers.write_once(offset, deadline as u32);
+    enable_machine_timer();
```
zevorn, line 172: [P2] Order mtimecmp before enabling MTIE

The final volatile MMIO store is immediately followed by a CSR write that enables MTIE. CSR accesses are weakly ordered relative to MMIO, so the old expired comparator may still be visible when MTIE becomes visible and cause an early/spurious timer interrupt. Insert `io_fence()` after the RV32-safe `mtimecmp` write sequence and before `enable_machine_timer()`, as the MSIP paths already do.

prototyper/machine/src/pmu/access.rs
```diff
@@ -0,0 +1,68 @@
...
+        NextMode::User => (TIME_COUNTER, TIME_COUNTER),
+        NextMode::Machine => (0, 0),
+    }
+}
+
+fn write_checked<const CSR: u16>(value: usize, mode: NextMode) -> Result<usize, CounterError> {
+    // SAFETY: callers instantiate only the fixed RISC-V counter-enable CSRs.
+    match unsafe { swap_csr::<CSR>(value) } {
+        ExpectedResult::Value(_) => match unsafe { probe_csr::<CSR>() } {
+            ExpectedResult::Value(actual) if actual == value => Ok(actual),
```
zevorn, line 40: [P2] Accept legal WARL hardwired-zero enable bits

`mcounteren` and `scounteren` are WARL and may legally read back a safe subset of the requested bits. Exact equality makes boot fail on a compliant hart that implements an M-mode counter but cannot expose it to S/U mode. Reject unexpected enabled bits, but retain/intersect the actual readable subset (including the supervisor mask) instead of requiring exact readback.

prototyper/machine/src/pmu/counters.rs
```diff
@@ -0,0 +1,201 @@
...
+        self.current_facts().map_or(0, |facts| facts.count())
+    }
+
+    /// Returns read-only facts for one dense current-hart counter index.
+    pub fn info(&self, index: usize) -> Result<CounterInfo, CounterError> {
+        let facts = self.current_facts()?;
+        let offset = facts.validate(index)?;
+        Ok(CounterInfo {
+            csr_number: supervisor_csr(offset),
+            width: 64,
```
zevorn, line 53: [P2] Report the probed width of each programmable counter

SBI `counter_get_info` exposes counter width because programmable counters may implement fewer than 64 bits. Hard-coding 64 produces incorrect wrap/delta calculations and conflicts with exact initial-value writes on such harts. Probe/store each counter's WARL precision and return that value here; cycle/instret can retain their architectural width.

prototyper/firmware/src/sbi/hsm.rs
```diff
@@ -0,0 +1,88 @@
...
+
+impl Hsm {
+    pub(super) fn new(harts: HartControl) -> Self {
+        Self { harts }
+    }
+}
+
+impl rustsbi::Hsm for Hsm {
+    fn hart_start(&self, hart_id: usize, start_addr: usize, opaque: usize) -> SbiRet {
+        let next_stage = match NextStage::supervisor(start_addr, opaque) {
```
zevorn, line 20: [P1] Validate HSM entry points against the executable PMP grant

`NextStage::supervisor` checks only the image contract's broad address interval and alignment, not whether S-mode has executable access under the PMP policy installed in `main`. An aligned address inside the RustSBI image is within the generated `0x8000_0000..0x9000_0000` interval, so `hart_start` can return success and then transfer the target hart to a PMP-forbidden address. HSM/SUSP require `INVALID_ADDRESS` when executable access is prohibited. Pass the selected supervisor executable range/capability into `Hsm` and validate start and non-retentive resume addresses before committing lifecycle state.

prototyper/firmware/src/device_tree/aia.rs
```diff
@@ -0,0 +1,423 @@
...
+        return Err(DiscoverError::UnsupportedDevice);
+    }
+    let default_hart_bits = topology_bits(raw_files.len())?;
+    let hart_index_bits = optional_u32(node, "riscv,hart-index-bits", default_hart_bits)?;
+    let group_index_bits = optional_u32(node, "riscv,group-index-bits", 0)?;
+    let group_index_shift = optional_u32(node, "riscv,group-index-shift", 24)?;
+    validate_topology(hart_index_bits, group_index_bits, group_index_shift)?;
+
+    let base = ranges.first().ok_or(DiscoverError::DeviceRange)?.start;
+    let supervisor_imsic_base = ranges
```
zevorn, line 130: [P1] Resolve the separate supervisor IMSIC node

Stock QEMU `virt,aia=aplic-imsic` emits a machine IMSIC at `0x2400_0000` and a separate supervisor IMSIC at `0x2800_0000`; each node has one `reg` tuple. Reading `ranges[1]` from the machine node therefore rejects the QEMU DT. Follow the machine APLIC's `riscv,children` and the child APLIC's `msi-parent` phandle to the supervisor IMSIC (and validate that node), then add the generated QEMU AIA DT as an integration fixture.

prototyper/firmware/src/device_tree/interrupts.rs
```diff
@@ -0,0 +1,69 @@
...
+use riscv_aia::Iid;
+
+pub(crate) fn install(
+    boot: &mut machine::BootInfo,
+    tree: &mut BootTree,
+    harts: &[usize],
+) -> Result<Option<machine::Interrupts>, Error> {
+    let (imsic, aplic) = aia::discover(tree, harts)?;
+    let clint = clint::discover(&tree.tree().root, harts)?;
+    if !description_is_coherent(imsic.is_some(), aplic.is_some(), clint.is_some()) {
```
zevorn, line 16: [P1] Accept the CLINT that coexists with QEMU AIA

QEMU's valid AIA `virt` DT contains both the APLIC/IMSIC hierarchy and `clint@2000000`. This coherence check rejects that stock description before installation, which explains why the AIA path is unreachable in the current boot job. When a complete AIA path is present, select its timer/notification mechanism and ignore or explicitly retire CLINT instead of treating coexistence as malformed; cover the actual QEMU DT.

prototyper/firmware/src/device_tree/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+                .property("device_type")
+                .and_then(|property| property.as_str().ok())
+                == Some("memory");
+        if !is_memory || !dt::enabled(node) {
+            continue;
+        }
+        let Some(reg) = node.property("reg") else {
+            continue;
+        };
+        let Some((start, size)) = dt::first_reg(reg.value(), address_cells, size_cells) else {
```
zevorn, line 63: [P1] Keep PMP policy consistent with every advertised RAM range

This returns only the first tuple of the first memory node, while the handed-off DT still advertises every other `reg` tuple and memory node. `main` consequently grants PMP/DBCN access only to this first range; a supervisor using another advertised bank will fault. Collect and validate all enabled RAM ranges, or reject/disable the ranges that are not delegated so the DT and authority handed to S-mode cannot disagree.

prototyper/firmware/src/sbi/pmu/event.rs
```diff
@@ -0,0 +1,102 @@
...
+
+impl Event {
+    pub(super) fn parse(index: usize, data: u64) -> Result<Self, Error> {
+        if index & !EVENT_INDEX_MASK != 0 {
+            return Err(Error::InvalidParam);
+        }
+        let kind = (index >> 16) & 0xf;
+        let code = index & 0xffff;
+        match kind {
+            event_type::HARDWARE_GENERAL
```
zevorn, line 31: [P1] Keep raw event_data out of firmware-owned control bits

For raw type #2, SBI gives the caller only 32 bits on RV32 or the low 48 bits on RV64; raw v2 gives 32/56 bits. The remaining upper bits are determined by the SBI implementation, but this parser accepts an arbitrary `u64` and forwards it unchanged to `mhpmevent`, letting S-mode control filter/overflow fields reserved to firmware. Validate/mask raw data by event type and XLEN and add boundary tests. See [SBI PMU raw event encodings](https://github.com/riscv-non-isa/riscv-sbi-doc/blob/master/src/ext-pmu.adoc).

prototyper/firmware/src/sbi/pmu/mod.rs
```diff
@@ -0,0 +1,582 @@
...
+        &self,
+        counter_idx_base: usize,
+        counter_idx_mask: usize,
+        start_flags: usize,
+        initial_value: u64,
+    ) -> SbiRet {
+        let Some(flags) = flags::CounterStartFlags::from_bits(start_flags) else {
+            return Error::InvalidParam.into();
+        };
+        if flags.contains(flags::CounterStartFlags::INIT_SNAPSHOT) {
```
zevorn, line 327: [P2] Enforce INIT_VALUE's single-counter contract

The PMU specification makes `INIT_VALUE` valid only for a single counter and mutually exclusive with `INIT_SNAPSHOT`. This implementation applies the same `initial_value` to every selected counter and does not validate the combination before processing snapshot availability. Validate both structural constraints before mutating counters and add multi-counter/combined-flag tests; do not let the loop silently extend a single-counter operation. See [counter-start semantics](https://github.com/riscv-non-isa/riscv-sbi-doc/blob/master/src/ext-pmu.adoc#function-start-a-set-of-counters-fid-3).

prototyper/firmware/src/sbi/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+            pmu.record(event);
+        }
+    }
+}
+
+impl machine::SbiHandler for Dispatcher {
+    fn handle(&self, call: SbiCall) -> sbi_spec::binary::SbiRet {
+        let result =
+            RustSBI::handle_ecall(self, call.extension_id, call.function_id, call.arguments);
+        self.record_sbi_call(call.extension_id, call.function_id);
```
zevorn, line 117: [P2] Do not count unavailable operations as completed telemetry

This records a sent/set PMU event solely from EID/FID even when the corresponding optional adapter is absent and dispatch returns `NOT_SUPPORTED`. For example, PMU can be installed while timer/IPI/RFENCE are all `None`, yet those counters still increment. That makes the advertised `SET_TIMER`/`IPI_SENT`/fence-sent telemetry describe calls the firmware could not perform. Move accounting into the adapter operation points (or explicitly document attempted-call semantics) and add tests for absent extensions and failed masks; the current test-kernel expectations also encode inconsistent behavior between running/stopped counters.

library/riscv-aia/src/aplic.rs
```diff
@@ -0,0 +1,443 @@
...
+    pub const fn new(
+        source_count: u32,
+        machine_imsic_base: u64,
+        supervisor_imsic_base: u64,
+        hart_index_width: u8,
+    ) -> Option<Self> {
+        if source_count == 0
+            || source_count > MAX_SOURCE_COUNT
+            || hart_index_width > 7
+            || !machine_imsic_base.is_multiple_of(0x1000)
```
zevorn, line 64: [P2] Reject IMSIC bases instead of silently rewriting them

`base_page_number` clears the low `hart_index_width` PPN bits after this API accepts any 4-KiB-aligned base. For example, `(0x2400_3000, width=2)` is programmed as `0x2400_0000`, so hart 0 is targeted outside the caller-supplied window. Under this no-group, 4-KiB-stride layout, reject bases not aligned to `1 << (12 + hart_index_width)` (and the analogous supervisor constraint), or model the complete HHX/LHX layout explicitly. Replace the current rounding test with a rejection test. See [AIA outgoing MSI addressing](https://github.com/riscv/riscv-aia/blob/20250312/src/AdvPLIC.adoc#addresses-and-data-for-outgoing-msis).

library/riscv-aia/src/aplic.rs
```diff
@@ -0,0 +1,443 @@
...
+        });
+    }
+
+    let domain = DomainConfig::from_raw(0)
+        .set_delivery_mode(1)
+        .set_big_endian(false)
+        .set_interrupt_enable(false)
+        .raw();
+    let expected_domain = DOMAINCFG_READ_ONLY | domain;
+    write(access, DOMAINCFG, expected_domain)?;
```
zevorn, line 152: [P2] Handle a legally preconfigured big-endian domain

A system reset leaves BE=0, but this library can also receive an APLIC retained or configured by previous firmware with `domaincfg.BE=1`. In that valid state, this first `0x8000_0004` write is byte-swapped, fails to establish DM=1, and the function rejects the device on readback. AIA recommends an endian-independent bootstrap write `(x << 24) | x` (here `x = 0x04`) before normal accesses. Implement that sequence or document/enforce a cold-reset-only precondition, and test a BE=1 backend. See [domaincfg initialization](https://github.com/riscv/riscv-aia/blob/20250312/src/AdvPLIC.adoc#domain-configuration-domaincfg).

prototyper/devkit/cargo-prototyper/src/artifact.rs
```diff
@@ -0,0 +1,186 @@
...
+            named_elf,
+            named_binary,
+            fit,
+        }
+    }
+}
+
+/// Executes Cargo and, for build plans, constructs the final image artifacts.
+pub fn execute(plan: &ExecutionPlan) -> Result<ArtifactSet, Error> {
+    prepare_link_inputs(plan)?;
```
zevorn, line 54: [P1] Make generated linker inputs part of Cargo's fingerprint

`prepare_link_inputs` rewrites fixed `payload.o`, `contract.o`, and DTB object paths, but their contents are outside Cargo's dependency graph and the `RUSTFLAGS` string does not change. A second build with a different payload/address/DTB is therefore reported fresh and reuses the previous linked bytes; this is reproducible by building the default payload and then passing another file. Use content-addressed object paths (so `RUSTFLAGS` changes) or another tracked/forced-relink mechanism, and test two consecutive builds with different inputs. Fixed paths also race across concurrent invocations.

prototyper/devkit/cargo-prototyper/src/project.rs
```diff
@@ -0,0 +1,77 @@
...
+            }
+        }
+        Err(Error::WorkspaceNotFound(absolute))
+    }
+
+    pub fn root(&self) -> &Path {
+        &self.root
+    }
+
+    pub fn target_dir(&self, target: &str, release: bool) -> PathBuf {
```
zevorn, line 36: [P1] Resolve the same target directory Cargo actually uses

The child Cargo honors `CARGO_TARGET_DIR` and Cargo config, but artifact discovery always reads `<workspace>/target`. With a custom target directory the build succeeds and this tool objcopies a stale root-tree ELF (or fails if none exists); this is reproducible with `CARGO_TARGET_DIR=<tmp>`. Use `cargo metadata`'s `target_directory` or pass one explicit `--target-dir` to the child and derive every artifact from it.

prototyper/devkit/cargo-prototyper/src/plan.rs
```diff
@@ -0,0 +1,543 @@
...
+    pub link_inputs: Vec<LinkInput>,
+    pub target_dir: PathBuf,
+    pub mode_suffix: Option<&'static str>,
+    pub firmware_type: Option<FirmwareType>,
+    pub pack: bool,
+}
+
+impl ImagePlan {
+    pub fn resolve(project: &Project, options: BuildOptions) -> Result<Self, Error> {
+        if options.pack && !matches!(options.role, ImageRole::Test | ImageRole::Bench) {
```
zevorn, line 152: [P1] Do not package RV32 with the RV64 FIT load address

This permits `--arch rv32 --image test --pack`, but the shared ITS fixes the kernel load to `0x8020_0000` while `test-rv32.ld` links/enters it at `0x8040_0000`. The command succeeds and emits an image that U-Boot loads at the wrong address. Generate the ITS from the resolved plan (including XLEN/link address), or reject RV32 packaging until an RV32 contract exists; add a `dumpimage` assertion.

prototyper/devkit/cargo-prototyper/src/artifact.rs
```diff
@@ -0,0 +1,186 @@
...
+    let relative = match plan.image.role {
+        ImageRole::Test => "prototyper/test-kernel/scripts/rustsbi-test-kernel.its",
+        ImageRole::Bench => "prototyper/bench-kernel/scripts/rustsbi-bench-kernel.its",
+        ImageRole::Firmware | ImageRole::Mtest => unreachable!("plan rejected this packaging role"),
+    };
+    let source = plan.project_root.join(relative);
+    if !source.is_file() {
+        return Err(Error::MissingInput(source));
+    }
+    let firmware = plan.image.target_dir.join("rustsbi-firmware.bin");
```
zevorn, line 146: [P1] Package a firmware artifact from the same resolved plan

`--pack` blindly consumes whichever generic `rustsbi-firmware.bin` was built last, without checking its architecture, firmware type, or load layout. The default firmware is FW_PAYLOAD and already contains the test kernel at `0x8020_0000`; the ITS then loads another kernel at the same address, and RV32/RV64 artifacts can also be mixed. Build or require a matching FW_DYNAMIC artifact in the same transaction and carry its provenance in `ExecutionPlan` instead of discovering a mutable generic filename.

prototyper/devkit/cargo-prototyper/src/main.rs
```diff
@@ -0,0 +1,440 @@
...
+    config.next_stage = match firmware_type {
+        FirmwareType::Dynamic => {
+            if arguments.next_stage_binary.is_some() || arguments.next_stage_address.is_some() {
+                return Err(cargo_prototyper::Error::InvalidFirmwareContract(
+                    "FW_DYNAMIC does not accept next-stage overrides",
+                ));
+            }
+            NextStageSource::Dynamic
+        }
+        FirmwareType::Jump => {
```
zevorn, line 321: [P1] Reject next-stage arguments that this firmware type ignores

FW_JUMP silently ignores `--next-stage-binary`, and FW_PAYLOAD silently ignores `--next-stage-address`; even a nonexistent ignored path exits successfully and can boot a different image than requested. Reject both conflicting combinations (the existing `ConflictingStageSources` error is currently unused) before resolving a mode, and add CLI tests for every pair of stage sources.

prototyper/devkit/cargo-prototyper/src/main.rs
```diff
@@ -0,0 +1,440 @@
...
+                    }
+                }
+            }
+        }
+    };
+    options.firmware = Some(config);
+    Ok(options)
+}
+
+fn parse_address(value: &str) -> Result<u64, cargo_prototyper::Error> {
```
zevorn, line 381: [P1] Validate the address against the generated firmware contract

This accepts values such as `0x2`, although `encode_contract` permits only `[0x8000_0000, 0x9000_0000)`, so the tool successfully produces an image that fail-stops at boot. It also accepts values above `u32::MAX` for RV32 and truncates/fails only in the firmware. Validate alignment, contract range, and selected XLEN here so every successful build has a usable next-stage address.

prototyper/devkit/cargo-prototyper/src/plan.rs
```diff
@@ -0,0 +1,543 @@
...
+            arguments.extend(["--".into(), "-D".into(), "warnings".into()]);
+        }
+
+        let mut environment = image.environment.clone();
+        environment.insert("RUSTFLAGS".into(), image.rustflags());
+        Ok(Self {
+            project_root: project.root().to_path_buf(),
+            action,
+            image,
+            program: "cargo".into(),
```
zevorn, line 355: [P1] Invoke the Cargo that launched this subcommand

Resolving plain `cargo` from `PATH` ignores Cargo's `CARGO` environment variable. An invocation through an explicit nightly Cargo can therefore launch stable Cargo internally and fail immediately on `-Z build-std`; this is reproducible even with `CARGO=<nightly cargo>` set. Use `std::env::var_os("CARGO")` with a `cargo` fallback, as `xtask/devkit.rs` already does, and add a forwarding test.

xtask/src/prototyper.rs
```diff
@@ -31,196 +28,29 @@ pub struct PrototyperArg {
...
-                .ok_or_else(|| format!("Invalid file path: {}", target))
-                .unwrap_or_else(|err| {
-                    eprintln!("Warning: {}. Falling back to target string.", err);
-                    target
-                })
-                .to_string()
-        } else {
-            target.to_string()
-        }
+    push_value(&mut forwarded, "--fdt", arguments.fdt.as_deref());
```
zevorn, line 37: [P1] Translate the legacy options before forwarding

This compatibility bridge sends retired flags (`--fdt`, `--payload`, `--target`, `--config-file`, and `--jump`) to a parser that accepts `--device-tree`, `--next-stage-binary`, `--arch`, `--config`, and `--firmware-type`. Clap rejects the current forwarding immediately, and `arceboot --payload` reaches this path. Translate values and mode semantics completely, including the old no-option FW_DYNAMIC default, and add forwarding tests.

xtask/src/arceboot.rs
```diff
@@ -100,7 +100,7 @@ pub fn run(arg: &ArcebootArg) -> Option<ExitStatus> {
             error!("Failed to build RustSBI Prototyper");
             return Some(exit_status);
         }
-        Some(target_dir.join("rustsbi-prototyper-payload.elf"))
+        Some(target_dir.join("rustsbi-firmware-payload.elf"))
```
zevorn, line 103: [P1] Use the devkit's actual firmware artifact

`target_dir` is ArceBoot's `riscv64gc` directory, while the new devkit defaults firmware to `riscv64imac` and names the ELF `rustsbi-firmware-fw-payload.elf`. After the forwarding bug is fixed this reconstructed path still does not exist. Return/use the devkit's `ArtifactSet`, or resolve the exact target and mode-specific name rather than deriving it from ArceBoot's output directory.

prototyper/devkit/cargo-prototyper/src/runner.rs
```diff
@@ -0,0 +1,305 @@
...
+            stderr,
+        },
+        Err(()) => Outcome::TimedOut { stdout, stderr },
+    })
+}
+
+fn read_all(mut stream: impl Read) -> Result<Vec<u8>, Error> {
+    let mut bytes = Vec::new();
+    stream
+        .read_to_end(&mut bytes)
```
zevorn, line 203: [P2] Bound captured guest output

QEMU controls these streams, so `read_to_end` grows memory without limit until exit or timeout; a broken/noisy guest can exhaust host or CI memory before the timeout is classified. Use a bounded ring buffer or explicit capture limit, preserve the useful tail, and report overflow in `Outcome`.

prototyper/bench-kernel/src/main.rs
```diff
@@ -205,19 +231,32 @@ extern "C" fn rust_main(hartid: usize, dtb_pa: usize) -> ! {
...
+            let property = node.get_prop("reg")?;
+            property
+                .deserialize::<Reg>()
+                .iter()
+                .next()
+                .map(|range| range.0.start)
+        })
+        .filter(|address| *address != 0)
+    {
+        UART.call_once(|| Uart16550Map(address));
```
zevorn, line 251: [P1] Do not access the UART after firmware retains it

The new firmware owns this UART: it marks the node disabled in the handed-off DT and its PMP policy grants S-mode only supervisor RAM, not QEMU's `0x1000_0000` MMIO window. Keeping a direct mapping here makes the first log write fault; a fresh FW_PAYLOAD bench run times out with no serial output, while test/mtest complete. Use SBI DBCN as test-kernel now does, or explicitly transfer both DT visibility and PMP authority.

prototyper/test-kernel/src/main.rs
```diff
@@ -104,16 +172,66 @@ extern "C" fn rust_main(hartid: usize, dtb_pa: usize) -> ! {
...
     let test_result = testing.test();
+    if !test_result {
+        protocol_record(|output| protocol::case_fail(output, test_run(), "sbi.core", "SBI_CORE"));
+        protocol_record(|output| protocol::run_fail(output, test_run(), 0, 1, "SBI_CORE"));
+        sbi::system_reset(sbi::Shutdown, sbi::SystemFailure);
+    }
+    protocol_record(|output| protocol::case_pass(output, test_run(), "sbi.core"));
 
+    protocol_record(|output| protocol::case_start(output, test_run(), "sbi.pmu"));
     pmu_test(smp);
```
zevorn, line 186: [P1] Keep the advertised RV32 suite from hanging in the PMU phase

This function is still documented as qemu-riscv64-only, but the new devkit unconditionally runs it for `cargo prototyper test --arch rv32`. Repeated RV32 runs enumerate all 34 counters and then hang here before RFENCE, SUSP, or any M-mode case; RV64 completes. Make the PMU sequence RV32-safe or gate it until supported, and add an RV32 end-to-end test that requires a terminal protocol record.

prototyper/test-kernel/src/main.rs
```diff
@@ -433,56 +556,28 @@ impl BoardInfo {
...
+    fn put_str(&self, mut value: &str) {
+        while !value.is_empty() {
+            let bytes = Physical::new(value.len(), value.as_ptr() as usize, 0);
+            let Some(written) = sbi::console_write(bytes).ok() else {
+                return;
+            };
+            if written == 0 {
+                return;
+            }
+            let Some(remaining) = value.get(written..) else {
```
zevorn, line 576: [P2] Preserve output after a partial UTF-8 write

SBI DBCN returns a byte count and may stop inside a multi-byte UTF-8 code point (the UART backend can make a one-byte partial write). `str::get(written..)` then returns `None` and silently drops every remaining byte. Track the remainder as `&[u8]` rather than slicing a `str` at an arbitrary byte offset, and test partial writes at each byte of a non-ASCII character.

prototyper/README.md
```diff
@@ -1,190 +1,130 @@
...
 ```
 
-This builds the firmware based on the provided options (or defaults if none are specified). The resulting files—such as `.elf` executables and `.bin` binaries—are generated in the `target/riscv64imac-unknown-none-elf/release/` directory under your project root. See the "Firmware Compilation" section for specific outputs and modes.
-
-These are necessary for building the firmware and handling RISC-V binary outputs.
+The most commonly used options are:
 
-#### Options
+- `-f, --features <FEATURES>` enables a comma-separated Cargo feature set.
+- `--fdt <PATH>` embeds the supplied flattened device tree.
```
zevorn, line 28: [P1] Regenerate these examples from the implemented CLI

The parser introduced here accepts `--device-tree`, `--next-stage-binary`, `--firmware-type FW_PAYLOAD|FW_JUMP`, `--next-stage-address`, `--config`, and `--arch rv32|rv64`; it rejects `--fdt`, `--payload`, `--jump`, `--config-file`, and `--target`. The no-subcommand shim does not translate the legacy names either. Rewrite this option list and every example below from `cargo prototyper --help`, and add CLI smoke tests so the primary README cannot drift back to commands that fail at argument parsing.

prototyper/README.md
```diff
@@ -1,190 +1,130 @@
...
-
-#### 1. Dynamic Firmware
-
-**Compilation Command:**  
-Use this command to compile firmware that dynamically loads payloads:
+### Dynamic image
 
 ```bash
-cargo prototyper
+cargo prototyper build
```
zevorn, line 43: [P1] Document the actual default mode and artifact names

A no-option build is FW_PAYLOAD, not dynamic: `firmware/Prototyper.toml` embeds `rustsbi-test-kernel`. It produces `rustsbi-firmware-fw-payload.{elf,bin}`; dynamic requires explicit `--firmware-type FW_DYNAMIC` and produces `rustsbi-firmware-fw-dynamic.{elf,bin}`. Update all three mode commands/name pairs and every boot guide that consumes the old `rustsbi-prototyper-*` artifacts.

prototyper/README.md
```diff
@@ -1,190 +1,130 @@
...
-- `heap_size`: Heap size, in bytes.
-- `page_size`: Page size, in bytes.
-- `log_level`: Logging level (`TRACE`, `DEBUG`, `INFO`, `WARN`, `ERROR`).
-- `jump_address`: Target address for jump mode.
-- `tlb_flush_limit`: TLB flush limit, in bytes.
+## Configuration
 
-To use a custom configuration file, specify it with:
+The default configuration is
+[`prototyper/config/default.toml`](prototyper/config/default.toml). A custom
```
zevorn, line 73: [P1] Describe the manifest that actually exists

This linked file does not exist, and the new schema does not expose the hart, memory, logging, jump-address, or TLB knobs described here. The implemented default is `firmware/Prototyper.toml`, selected with `--config`, and currently contains only `platform`, `firmware_type`, `device_tree`, and `[next_stage]`. Update the link, flag, and configuration contract together.

prototyper/README.md
```diff
@@ -1,190 +1,130 @@
...
-A tool to run code checks before committing:
+Run the ordinary Rust checks before submitting a change:
 
 ```bash
-pipx install pre-commit
-pre-commit install  # Set up pre-commit for the project
+cargo fmt --all -- --check
+cargo test -p cargo-prototyper
+cargo test -p rustsbi-prototyper-machine
+cargo test -p rustsbi-prototyper
```
zevorn, line 106: [P2] List development checks that run from a clean checkout

`rustsbi-prototyper` was removed, so this command fails with “package ID specification did not match any packages” and the current CI fails on the same line. The preceding host `cargo test -p rustsbi-prototyper-machine` also fails on non-RISC-V hosts because its test build names fixed RISC-V registers. Replace this block and the workflow with the actual host-runnable tests plus target-specific RV32/RV64 checks.

prototyper/devkit/README.md
```diff
@@ -0,0 +1,30 @@
...
+QEMU 参数，目标代码不会依赖这个工具。
+
+常用命令：
+
+```console
+cargo prototyper build
+cargo prototyper build --image mtest
+cargo prototyper build --image test
+cargo prototyper build --image bench
+cargo prototyper check --target riscv32imac-unknown-none-elf
```
zevorn, line 14: [P1] Keep this quick start synchronized with --help

The parser has no `--target`; architecture selection is `--arch rv32` or `--arch rv64`. The custom-payload example on line 27 is also rejected: the implemented form is `--firmware-type FW_PAYLOAD --next-stage-binary <BIN>`. Update both examples and cover them with CLI smoke tests.

prototyper/devkit/README.md
```diff
@@ -0,0 +1,30 @@
...
+语义，不等同于当前临时保留的 Cargo package 名。
+
+普通 `cargo check` 仍可用于快速检查 Rust 源码；只有
+`cargo prototyper build` 承诺生成采用受审链接布局、完成 ELF 到裸二进制转换并
+使用稳定名称的可启动最终镜像。
+
+不带额外参数的 `run` 会自动构建默认 S-mode 测试 payload、嵌入固件并启动
+QEMU。只有运行自定义 payload 时才需要传 `--payload <BIN>`。
+
+旧的 `cargo xtask prototyper`、`cargo xtask test` 和
```
zevorn, line 29: [P1] Do not promise a compatibility shim that rejects legacy input

`cargo xtask prototyper` forwards `--jump`, `--payload`, `--fdt`, and target triples unchanged to a parser that rejects them, and its no-option meaning changed from dynamic to payload. Either translate the legacy interface and defaults completely or remove/defer this compatibility claim.

prototyper/docs/booting-fedora-in-qemu-using-uboot-and-rustsbi.md
```diff
@@ -61,23 +62,24 @@ $ cargo prototyper
...
+$ cd "$WORKSHOP_DIR/u-boot"
 ```
 
 导出环境变量
 
 ```shell
 $ export ARCH=riscv
 $ export CROSS_COMPILE=riscv64-linux-gnu-
-$ export OPENSBI=../rustsbi/target/riscv64gc-unknown-none-elf/release/rustsbi-prototyper.bin
+$ export OPENSBI="$WORKSHOP_DIR/rustsbi/target/riscv64gc-unknown-none-elf/release/rustsbi-prototyper.bin"
```
zevorn, line 73: [P1] Build and consume the dynamic artifact produced by this PR

This edited path cannot be produced by the new devkit: the target is `riscv64imac-unknown-none-elf` and the binary is `rustsbi-firmware-fw-dynamic.bin`. Bare `cargo prototyper` now builds FW_PAYLOAD, while U-Boot's `OPENSBI` input requires a dynamic handoff image. Use `cargo prototyper build --firmware-type FW_DYNAMIC` and the matching output here, then apply the same migration to the FreeBSD, Linux, openEuler, OpenWrt, PolyOS, test-kernel, Ubuntu, and EDK2 guides.

prototyper/docs/booting-in-visionfive2-using-uboot-and-rustsbi.md
```diff
@@ -82,16 +88,16 @@ $ git clone https://github.com/rustsbi/rustsbi
...
 ```
 
 
 编译 RustSBI Prototyper，以 U-Boot 作为 Payload。
 
 ``` shell
-$ cd workshop/VisionFive2/rustsbi 
-$ cargo prototyper --payload ../work/u-boot/u-boot.bin --fdt ../work/u-boot/arch/riscv/dts/starfive_visionfive2.dtb 
+$ cd "$WORKSHOP_DIR/VisionFive2/rustsbi"
+$ cargo prototyper --target riscv64imac-unknown-none-elf --payload ../work/u-boot/u-boot.bin --fdt ../work/u-boot/arch/riscv/dts/starfive_visionfive2.dtb
```
zevorn, line 100: [P1] Do not document a platform this devkit rejects

This command uses three nonexistent options. Merely renaming them is insufficient: manifest validation currently accepts only `platform = "qemu-virt"`, so this PR has no supported VisionFive2 platform/link profile. Add the platform support and its checked link contract before documenting this flow, or clearly remove/defer the guide and its stale artifact path.

prototyper/docs/firmware-internal-privilege-separation.md
```diff
@@ -0,0 +1,45 @@
...
+这种组织方式的直接问题并非只有 `unsafe` 代码数量。一个处理 SBI 语义的模块若
+可以取得任意 MMIO 地址、通用 CSR 或可写的 trap frame，它也就同时拥有了改变
+机器执行状态的能力。随着计时器、IPI、远程 fence、电源管理、性能计数器和多个
+hart 的支持逐步加入，权限会沿着方便的接口扩散，最终使得一个很小的策略修改也
+不得不理解启动、异常恢复和物理内存的全局不变量。
+
+RustSBI Prototyper 将这种问题称为**固件内特权分离架构**。这里的“固件内”很
+重要：分离发生在同一份 M 态固件内部，而不是以另一种方式重复 RISC-V 的 M/S
+硬件特权级。`machine` 层持有直接作用于机器状态的机制权力；上层
+`rustsbi-prototyper` 则以安全 Rust 编写平台策略和 SBI 服务。上层能够表达“启动
```
zevorn, line 23: [P2] Use the interface names readers can find in this implementation

The upper package is `rustsbi-firmware`, not `rustsbi-prototyper`, and the handler returns `sbi_spec::binary::SbiRet`, not `SbiResponse` as stated later. Update both names so this new boundary document maps back to real modules and types.

.cargo/config.toml
```diff
@@ -1,13 +1,10 @@
 [alias]
 xtask = "run --package xtask --release --"
 arceboot = "xtask arceboot"
-prototyper = "xtask prototyper"
+prototyper = "run --package cargo-prototyper --"
```
zevorn, line 4: [P1] Migrate every consumer before switching the repository alias

This redirects all existing callers to the new parser, but the workflows and scripts still invoke `--jump`/`--payload`, test the deleted package, and consume `riscv64gc/.../rustsbi-prototyper*`; those jobs now fail before exercising firmware. Either retain a translating frontend at this switch point or update all CI/scripts and boot consumers in the same change.

library/sbi-testing/src/thread.rs
```diff
@@ -97,6 +97,24 @@ impl Thread {
                 sstatus       = inlateout(reg) sstatus,
                 execute_naked = sym execute_naked,
             );
+            #[cfg(target_pointer_width = "32")]
+            core::arch::asm!(
+                "   csrw sscratch, {sscratch}
+                csrw sepc    , {sepc}
+                csrw sstatus , {sstatus}
+                addi sp, sp, -4
```
zevorn, line 105: [P1] Preserve the declared C ABI's stack alignment

`execute_naked` is declared `extern "C"`; `#[naked]` suppresses compiler-generated prologue/epilogue but does not waive that calling convention. The RISC-V psABI requires `sp` to be 16-byte aligned on procedure entry, while this leaves RV32 at 12 mod 16 before `call` (and the RV64 sibling at 8 mod 16). Reserve a 16-byte save area on both targets, store `ra` in one slot, and leave padding so the naked callee enters with an ABI-compliant stack; otherwise expose/document an explicitly custom assembly convention instead. See [The Stack](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc#the-stack).

zevorn (changes requested):
Requesting changes on `0e0546c84dcb9d0faec6b44602a15baad995c3ab`: this refactor still has boot, concurrency, SBI/AIA-conformance, build-reproducibility, and migration blockers.

Coverage: I reviewed all 176 Rust paths in the fixed base…head diff (17,920 right-side Rust additions across 138 files) and all 15 changed Markdown documents. The review applied the [Asterinas coding guidelines](https://github.com/asterinas/asterinas/tree/main/book/src/to-contribute/coding-guidelines)—especially boundary validation, checked arithmetic, type invariants, concurrency, resource bounds, and hardware/ABI rules—together with the official SBI PMU/HSM, AIA, RISC-V psABI, and RustSBI contribution requirements. The inline comments below are the actionable, deduplicated results after a separate false-positive audit.

Two cross-cutting blockers have no changed RIGHT-side line on which GitHub can anchor a comment:

- `.github/workflows/prototyper.yml`, `.github/workflows/arceboot.yml`, `.github/scripts/prototyper-qemu-boot.sh`, and `.github/scripts/prototyper-openeuler-aia.sh` still invoke retired flags, test the deleted `rustsbi-prototyper` package, or consume `riscv64gc/.../rustsbi-prototyper*`. They must be migrated with the CLI/artifact switch; the current Prototyper, ArceBoot integration, and AIA jobs fail before providing the intended coverage.
- Public changes in `riscv-aia`, `sbi-rt`, `sbi-spec`, and `sbi-testing` have no `[Unreleased]` entries. This conflicts with the [RustSBI contribution guide](https://github.com/rustsbi/slides/blob/main/2025/reports/Contributing%20to%20RustSBI.md) and is already caught by the changelog check.

Validation included `cargo fmt --all -- --check`, targeted host tests for `riscv-aia`, `sbi-spec`, `sbi-rt`, and `cargo-prototyper`, RV64 QEMU test/mtest runs, generated-QEMU-DT inspection, and reproductions of stale relinking, custom-target-dir artifact selection, the RV32 test hang, and the bench UART/PMP timeout.

luojia65:
I propose to split this PR into independent modules: https://github.com/rustsbi/rustsbi/issues/219

PR #220, opened 2026-08-04
fix(aia): refine `riscv-aia` registers

luojia65:
- Align Smaia, Ssaia, and H/VS CSR masks and fields with RISC-V AIA 1.0.
- Fix `*topei`, `mvien`/`mvip`, VS interrupt shifts, and missing overflow/RAS fields.
- Add priority/IID setters, remove redundant accessors, and separate `Iid` from `MajorIid`.
- Improve one-hot and boundary tests.

Note to maintainter: Add `Signed-off-by: Zhenchen Wang <plucky@openatom.club>` on squash and merge. @Plucky923 

<details>

</details>


library/riscv-aia/src/register/hideleg.rs
```diff
@@ -1,9 +1,18 @@
...
+// > hideleg may be writable or may be read-only zero, depending on the implementation.
+//
+// Bits 13-63 of `hideleg` is writable in some of the implementations.
 riscv::read_write_csr! {
     /// Hypervisor interrupt delegation.
     Hideleg: 0x603,
-    mask: 0x444,
+    // 0xFFFF_E444 in RV32, or 0xFFFF_FFFF_FFFF_E444 in RV64
+    mask: usize::MAX & !0x1BBB,
 }
```
Plucky923, line 9: I noticed that similar files and similar code definitions do not have this many comments referencing the specification.

library/riscv-aia/src/register/stopei.rs
```diff
@@ -1,25 +1,25 @@
 //! Supervisor top external interrupt (only with an IMSIC) (stopei)
 
-use crate::Iid;
+use crate::iid::Iid;
 
-riscv::read_only_csr! {
+riscv::read_write_csr! {
```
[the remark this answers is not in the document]
luojia65: Writes to `*stopei` have special meanings which means claiming the current interrupt, `read_only` is not compatible.

library/riscv-aia/src/register/hideleg.rs
luojia65: Fixed in https://github.com/rustsbi/rustsbi/pull/220/commits/5d151e10ea16fef13d167b794574b507f6204a5b

library/riscv-aia/src/register.rs
```diff
@@ -21,15 +17,11 @@ pub mod mvip;
 pub mod midelegh;
 pub mod mieh;
 pub mod miph;
-pub mod mvieh;
+pub mod mvienh;
```
Plucky923, line 20: Could we gate the high-half CSR modules and accessors to RV32?

These CSRs are documented as RV32-only, but `pub mod mvienh;` is exported unconditionally. Moreover, `riscv::read_write_csr!` enables both `riscv32` and `riscv64` targets by default. As a result, an RV64 consumer can call `mvienh::read()` and generate a CSR access to `0x318`. On a compliant RV64 implementation, this access is not valid and may raise an illegal-instruction exception at runtime.`cargo check --target riscv64gc-unknown-none-elf` would still pass because it only validates compilation; it does not execute the CSR instruction.
Could we either:
1. add `#[cfg(target_arch = "riscv32")]` to the high-half modules and re-exports; or
2. use the `*_rv32` variants for all read/write/set/clear CSR macros?

I noticed that some of this behavior already existed before this PR. However, this PR introduces the `mvienh` path and adds more high-half CSR accessors, so it would be good to address it here.

library/riscv-aia/src/register/mtopei.rs
```diff
@@ -4,26 +4,26 @@
...
-use crate::Iid;
+use crate::iid::Iid;
 
-riscv::read_only_csr! {
+riscv::read_write_csr! {
     /// Machine top external interrupt (mtopei).
     Mtopei: 0x35C,
-    mask: 0x0FFF_00FF,
+    mask: 0x07FF_07FF,
 }
```
Plucky923, line 13: Could we avoid exposing `mtopei` through the generic `read_write_csr!` API?

The field layout and the new `0x07FF_07FF` mask look correct. However, `mtopei` has special claim semantics: a write clears the pending bit of the currently reported interrupt, and the value written by software is ignored.

The AIA specification requires the interrupt ID to be read and claimed atomically, for example with:
```
 csrrw rd, mtopei, x0
```
With the current API, callers can easily write:
```
let interrupt = mtopei::read();
mtopei::write(interrupt);
```
If a higher-priority interrupt becomes pending between these two instructions, the subsequent write may clear the new interrupt instead of the one returned by `read()`, causing the new interrupt to be lost.
Could we provide a dedicated API implemented as one atomic CSR instruction.

library/riscv-aia/README.md
Plucky923, line 1: Could we avoid maintaining the same crate introduction independently in both`README.md` and `src/lib.rs`?

Although `Cargo.toml` declares `readme = "README.md"`, rustdoc/docs.rs renders the crate-level documentation from `src/lib.rs`; it does not automatically include the  README content. The PR updates the specification version in both places, but they can drift in future changes.
Could we either use:
```
#![doc = include_str!("../README.md")]
```
if the README should be the single source of truth, or keep a deliberately short crate-level overview and document that the two files are maintained separately?

luojia65: Fixed in https://github.com/rustsbi/rustsbi/pull/220/commits/3d1bae50b5f2bffac34ad94cd3f84cb6c7218bb0

library/riscv-aia/src/register.rs
luojia65: Fixed in https://github.com/rustsbi/rustsbi/pull/220/commits/ba174c6437fe7ad3cc78f193fc1b4d2a54c15122 . I suggest to gate read/write functions other than the structures so that future RISC-V emulators will use them regardless of host architecture.

library/riscv-aia/src/register/mtopei.rs
luojia65: Marked read-only, https://github.com/rustsbi/rustsbi/pull/220/commits/35a9118fe5b858b376a3c17201ee58805343e109

library/riscv-aia/src/register/mtopi.rs
```diff
@@ -12,9 +12,12 @@ riscv::read_only_csr! {
...
     #[inline]
-    pub const fn iid(self) -> Option<Iid> {
-        let bits = (self.bits & 0x0FFF_0000) >> 16;
-        Iid::new(bits as u16)
+    pub const fn iid(self) -> Option<MajorIid> {
+        let number = ((self.bits & 0x0FFF_0000) >> 16) as u16;
+        match number {
+            0 => None,
+            _ => Some(MajorIid::new(number)),
+        }
```
Plucky923, line 22: `MajorIid` allows IID 0, but `mtopi`, `stopi`, and `vstopi` still return `None` whenever the decoded IID is zero.
For example, `0x0000_0001` has IID 0 and IPRIO 1, so the CSR is nonzero and represents an interrupt. The absence check should be based on the complete masked CSR value, not only the IID field.
Please return `None` only when the whole CSR is zero, and add a regression test for `from_bits(0x01)`. The same fix is needed in all three `*topi` registers.

 Use a more descriptive name than `number`

`number` is ambiguous here. It actually represents the decoded major interrupt identity (IID), not an arbitrary number.
Please consider renaming it to `iid` or `major_iid` to make the code easier to understand.

library/riscv-aia/src/register/hviprio1.rs
```diff
@@ -1,38 +1,148 @@
...
         ((self.bits >> shift) & 0xFF) as u8
     }
+
+    /// Sets the priority byte at byte index `i`.
+    #[inline]
+    const fn set_prio_byte(&mut self, i: usize, value: u8) {
+        let shift = i * 8;
+        self.bits = (self.bits & !(0xFFusize << shift)) | ((value as usize) << shift);
+    }
 }
```
Plucky923: Use clearer names for the byte helpers.`prio_byte` and `set_prio_byte` are somewhat ambiguous, and `i` does not indicate that it is a byte index rather than an interrupt number.

Consider names  to make the packed-register layout easier to understand.

library/riscv-aia/src/register/hideleg.rs
```diff
@@ -24,6 +26,26 @@ riscv::read_write_csr_field! {
...
 
+riscv::read_write_csr_field! {
+    Hideleg,
+    /// Counter overflow interrupt delegation.
+    counter_overflow: 13,
+}
+
+#[cfg(not(target_pointer_width = "32"))]
+riscv::read_write_csr_field! {
```
Plucky923, line 36: `not(target_pointer_width = "32")` is being used as a proxy for RV64, but it also matches other 64-bit architectures such as `x86_64`.

If this condition is about the CSR being RV64-specific, please use `target_arch ="riscv64"`. If it is only protecting accesses to bits above 31, `target_pointer_width= "64"` would make the intent clearer.

library/riscv-aia/src/register/mtopi.rs
luojia65: Fixed at https://github.com/rustsbi/rustsbi/pull/220/commits/ea1430473a60442987b113bde68bba7908e842e3

library/riscv-aia/src/register/hviprio1.rs
luojia65: Fixed at https://github.com/rustsbi/rustsbi/pull/220/commits/ea1430473a60442987b113bde68bba7908e842e3

library/riscv-aia/src/register/hideleg.rs
luojia65: Fixed at https://github.com/rustsbi/rustsbi/pull/220/commits/ea1430473a60442987b113bde68bba7908e842e3

Plucky923:
[fix(riscv-aia): apply review suggestions; add docs.rs package metadata](https://github.com/rustsbi/rustsbi/pull/220/commits/ea1430473a60442987b113bde68bba7908e842e3)

Commit messages cannot contain language like "apply review suggestions." Each commit should be atomic and precisely describe what was done.

library/riscv-aia/src/register/mtopei.rs
Plucky923: The problem. This PR currently models mtopei, stopei, and vstopei using read_only_csr!. However, *topei CSRs have an important side effect defined by the AIA 1.0 specification: writing the CSR performs the claim operation for the interrupt reported by the CSR. The written value is ignored, but the write clears the pending bit of the corresponding interrupt in the IMSIC interrupt file.

Therefore, treating *topei as read-only leaves out an essential part of its architectural semantics. An external-interrupt handler must eventually claim the reported interrupt; otherwise, the handler cannot correctly complete interrupt processing.

More importantly, the specification requires the read and claim to be performed atomically by a single CSR instruction. If the software first reads *topei and later writes it using a separate instruction, another higher-priority interrupt may become pending between the two operations. In that case, the write may claim the newly arrived interrupt rather than the one returned by the earlier read, causing the new interrupt to be lost.

For this reason, exposing ordinary read() and write() operations would also be the wrong abstraction. The API should represent the architectural operation directly: read the current top external interrupt and claim that same interrupt atomically.

This limitation is already visible in the repository. Since the CSR crate currently provides no claim operation, prototyper implements its own mtopei_claim in prototyper/prototyper/src/platform/aia.rs using a raw csrrw 0x35C, followed by a separate local decoding of the returned value.

That duplicated decoding has already diverged from the implementation in this PR: prototyper uses the legacy 12-bit IID mask 0x0FFF_0000, while this PR correctly changes the IID field to 11 bits with 0x07FF_0000. This does not currently cause an observable failure because the specification guarantees that bit 27 reads as zero, but it demonstrates the maintenance problem: the same architectural CSR is now implemented and decoded independently in two places, and the definitions have already drifted.

Suggested design. Rather than making Mtopei, Stopei, and Vstopei generally readable and writable, I suggest adding a dedicated claim() operation:

claim() executes a single CSRRW instruction with x0 as the source register, so reading the top interrupt and clearing its pending bit are one architectural operation.
It returns Option<Iid>, with None representing the zero value defined by the specification when no pending-and-enabled interrupt exists.
The returned value is decoded through the same 11-bit IID representation introduced by this PR.
No generic write() API is exposed, so callers cannot accidentally express the unsafe two-instruction read()-then-write() sequence.

For example:
```rust
impl Mtopei {
    /// Atomically reads and claims the top external interrupt.
    #[inline]
    pub fn claim() -> Option<Iid> {
        let bits: usize;
        unsafe {
            core::arch::asm!(
                "csrrw {bits}, 0x35C, zero",
                bits = out(reg) bits,
            )
        };
        Self::from_bits(bits).iid()
    }
}
```

library/riscv-aia/src/register/sie.rs
```diff
@@ -12,44 +13,93 @@ impl Sie {
     pub const fn bit(self, n: usize) -> bool {
         ((self.bits >> n) & 1) != 0
```
Plucky923: The bounds assert is the right contract — but the same `bit()` helper on `Sie`/`Sip`/`Vsie`/`Vsip` (e.g. `sie.rs:13`) kept the
 unguarded form, so the eight sibling APIs now disagree: M/MV-level panics on out-of-range, S/VS-level silently wraps in release builds.
 Could we add the same assert to the S/VS-level helpers too, so all eight share one contract?

library/riscv-aia/src/register/hviprio1.rs
Plucky923: The rename to `priority_at_byte_index` here looks good — but the same helpers in the three sibling files kept the old names:
 `prio_byte`/`set_prio_byte` with parameter `i` in `hviprio2.rs:118`, `hviprio1h.rs:49`, and `hviprio2h.rs`. Since all four model the same packed-priority layout, could we apply the rename there too?

library/riscv-aia/src/register/hideleg.rs
Plucky923: One leftover: the sibling registers (`hvien.rs:17`, `hvip.rs`, `mie.rs`, `mip.rs`, `mvien.rs`, `mvip.rs`, `mideleg.rs`, `hviprio1.rs`) still use `#[cfg(not(target_pointer_width = "32"))]` for the same purpose. Both forms behave identically on riscv32/riscv64, but having the two side by side reads as if the difference were meaningful. Could we standardize crate-wide on one form (either is fine)?

Plucky923:
> Note to maintainter: Add `Signed-off-by: Zhenchen Wang <plucky@openatom.club>` on squash and merge. @Plucky923

If you want to add them, please add all of them. Please keep all commit messages consistent, or you may choose not to add them.

Plucky923:
Similar tests currently use several different naming patterns across the new CSR files. For example, RAS one-hot checks are named *_ras_fields_are_one_hot in hideleg/hvien/mvien, but *_ras_fields in hvip/mie/mip. Likewise, field-decoding tests use suffixes such as *_parsing, *_value_fields, *_low_fields, and *_edge_cases for essentially the same kind of check.

Since this PR introduces a substantial set of new CSR tests, it would be useful to establish a small, consistent vocabulary for the aspect being tested, e.g. <register>_<fields_one_hot|mask|parse|set_get|zero|bounds>. This would make related tests easier to identify and keep naming consistent as the crate grows.

luojia65:
> > Note to maintainter: Add `Signed-off-by: Zhenchen Wang <plucky@openatom.club>` on squash and merge. @Plucky923
> 
> If you want to add them, please add all of them. Please keep all commit messages consistent, or you may choose not to add them.

There are too much commits in this PR. All the commit messages will be merged in final squash and merge, with all the Signed-off-by merged.

Plucky923 (approved):
LGTM! Thanks for your contribution!

zevorn (commented):
I think we should use more detailed commit bodies to explain why a patch is needed and why it's being introduced. This will improve overall readability and make it easier for people from different domains to cross-review the code.

zevorn (commented):
> fix(aia): add missing functions in hvip and hviph registers

I believe the commit is a feature patch, so using the fix prefix module can't... Well, maybe it's not the best, or is that just the development style of RustSBI?

luojia65:
@zevorn This is a feature patch, the commit message will be edited when squash and merge.

zevorn:
> @zevorn This is a feature patch, the commit message will be edited when squash and merge.

Thanks for the clear, LGTM

zevorn (approved):

PR #221, opened 2026-08-11
refactor(sbi-testing): split thread context switch per architecture 

Plucky923:
Restructure the `sbi-testing` thread context switch into a shared module plus per-architecture switchers, and re-enable the RV32 build in CI.

   ## What

   - Split `thread.rs` into `thread/mod.rs` (shared `Thread` context and safe API) + `thread/rv64.rs` / `thread/rv32.rs` (per-arch naked
 switchers, 8/4-byte register slots). RV32 support is new; the RV64 protocol is unchanged.
   - Pin the `Thread` memory layout against the assembly with compile-time `offset_of!` assertions — a field reorder now fails the build
 instead of silently restoring wrong registers.
   - Take the stack pointer in `Thread::new(entry, sp)` instead of a separate mutation; debug-assert register indices.
   - Keep the `call` site 16-byte ABI-aligned when saving `ra`.
   - English docs with the layout contract spelled out; callers (`spi.rs`, `time.rs`) adapted, no behavior change.
   - CI: restore `riscv32imac-unknown-none-elf` in the build-sbi-testing matrix.

   ## Verification

   - `cargo build -p sbi-testing` passes for riscv64imac and riscv32imac, default features and `--features log`.
   - The naked switcher bodies are byte-identical to the versions previously validated on qemu (RV64 prototyper boot and RV32 OpenSBI boot); the only delta is the ra-frame alignment fix, which will be exercised at the next qemu boot run.

   ## Notes

   - The RV32 runtime boot test is intentionally not included; it returns with the RV32 redo after the planned sbi-testing consolidation.
   - Changelog entries updated per repo conventions (`check-changelog` gate satisfied).
<details>

</details>


luojia65 (approved):
LGTM

PR #223, opened 2026-08-15
refactor(prototyper): separate entry mechanism from policy

Plucky923:
Add the entry-macros crate whose #[entry] attribute owns everything around the policy main: it generates the _start entry with the startup ceremony, the relocation subroutine, and the extern "C" bridge into policy code, so the policy main becomes an ordinary safe Rust function. The assembly itself lives in standalone src/entry/*.S files, included by the macro, and the ceremony steps are documented there.

<details>

</details>


prototyper/macros/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "macros"
```
luojia65:
```suggestion
name = "rustsbi-prototyper-macros"
```

prototyper/prototyper/src/main.rs
```diff
@@ -34,18 +34,16 @@ use crate::sbi::hart_context::NextStage;
 use crate::sbi::heap::sbi_heap_init;
 use crate::sbi::hsm::local_remote_hsm;
 use crate::sbi::ipi;
-use crate::sbi::trap;
 use crate::sbi::trap_stack;
-
-pub const R_RISCV_RELATIVE: usize = 3;
+use ::macros::entry;
```
luojia65:
```suggestion
use rustsbi_prototyper_macros::entry;
```

prototyper/macros/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+    let Ok(policy) = syn::parse2::<ItemFn>(item) else {
+        return quote! { compile_error!("`entry` can be applied only to a function"); };
+    };
+    let name = &policy.sig.ident;
+
+    quote! {
+        #policy
+
+        const _: () = {
+            /// Connects the startup ceremony to the policy function.
```
luojia65, line 30: This creates an anonymous namespace which looks great.

.github/workflows/prototyper.yml
```diff
@@ -31,7 +31,7 @@ jobs:
 
       - name: Run test
         run: |
-            cargo test -p rustsbi-prototyper
+            cargo test -p rustsbi-prototyper -p macros
```
luojia65:
```suggestion
            cargo test -p rustsbi-prototyper -p rustsbi-prototyper-macros
```

prototyper/prototyper/Cargo.toml
```diff
@@ -36,6 +36,7 @@ pastey = "0.1.0"
 uart_sifive = { git = "https://github.com/duskmoon314/uart-rs/" }
 arm-pl011-uart = "0.3.2"
 riscv-aia = { version = "0.0.0", path = "../../library/riscv-aia" }
+macros = { version = "0.0.0", path = "../macros" }
```
luojia65:
```suggestion
rustsbi-prototyper-macros = { version = "0.0.0", path = "../macros" }
```
Note that the folder name in `path` (in this case, `macros`) does not need to be the same as its crate name (`rustsbi-prototyper-macros`).

luojia65 (changes requested):
Rename the crate by adding a prefix, otherwise on crates.io publish, another crate named `macro` will collapse with this name, leading to publish failure. Or if this crate is not designed to be published from the beginning, add `publish = false` in Cargo.toml configuration file. Other parts LGTM including assembly code.

prototyper/prototyper/src/main.rs
```diff
@@ -34,18 +34,16 @@ use crate::sbi::hart_context::NextStage;
...
 
 #[inline(always)]
 fn has_mstateen0() -> bool {
     has_csr!(CSR_MSTATEEN0)
 }
 
-#[unsafe(no_mangle)]
-extern "C" fn rust_main(_hart_id: usize, opaque: usize, nonstandard_a2: usize) {
+#[entry]
+fn rust_main(_hart_id: usize, opaque: usize, nonstandard_a2: usize) {
```
luojia65: This function already looks great. It will be even better if we do a simple rename:

```suggestion
fn main(_hart_id: usize, opaque: usize, nonstandard_a2: usize) {
```

Now it follows the convention for naming every entry function as `fn main`. The name doesn't matter here technically, as the `#[entry]` macro routes them to `_start` whatever the name is. I suggest `fn main` because it looks ideal for naming convention rules, but not for technological limits.

luojia65 (approved):

PR #225, opened 2026-08-15
fix(workflow): filter CI triggers and cache EDK2 build

### LINKED ISSUE #224: CI: ArceBoot workflow runs on unrelated changes and its integration tests are slow
## Problem

The ArceBoot CI workflow (`.github/workflows/arceboot.yml`) currently triggers on **every** pull request and almost every push:

```yaml
on:
  pull_request:            # no paths filter at all
  push:
    paths-ignore:          # only docs are excluded
      - '**.md'
      - 'LICENSE'
```

As a result, PRs that touch code completely unrelated to ArceBoot (e.g. `library/sbi-spec`, `library/macros`, docs, other workflows) still run the full ArceBoot CI.

Additionally, the `integration-test` job is very slow:

- It builds **EDK2 from source** (`arceboot/scripts/test/build_edk2.sh`), which dominates the job runtime.
- It then generates a disk image and runs **three QEMU boot tests** (HelloRiscv, AllocatePage, Hello), each with `timeout 60`.

This slows down PR feedback and consumes CI minutes on unrelated changes.

## Prior art in this repo

`.github/workflows/openeuler-aia.yml` already uses explicit `paths:` filters on both `pull_request` and `push`, so it only runs when relevant files change.

## Proposals

1. **Add `paths:` filters to `arceboot.yml`** (both `pull_request` and `push`):
   - `build-and-clippy` (ArceBoot itself) only depends on: `arceboot/**`, `xtask/**`, `.github/workflows/arceboot.yml`, `Cargo.toml`, `Cargo.lock`, `rust-toolchain.toml`
   - `integration-test` boots `rustsbi-prototyper-payload.elf` as `-bios`, so it is additionally affected by: `prototyper/**`, `library/rustsbi/**`, `library/sbi-rt/**`, `library/sbi-spec/**`

   Since `on:` paths are workflow-level, possible implementations:
   - use the union of both path lists for the whole workflow (simplest), or
   - split into two workflow files: fast `arceboot.yml` (build + clippy) and heavy `arceboot-integration.yml` (QEMU tests), or
   - keep one workflow and gate the `integration-test` job with a path-detection step (e.g. `dorny/paths-filter`).
2. **Cache the EDK2 build** with `actions/cache`, keyed on the EDK2 source/submodule revision — it only needs rebuilding when its inputs change.
3. **Add `workflow_dispatch:`** so the heavy integration tests can still be triggered manually at any time.
4. (Optional) If per-PR integration coverage is not required, run the QEMU integration tests only on pushes to `main` and/or on a nightly schedule.

jack-wang-176:
Only run ArceBoot CI when relevant paths change, so unrelated modifications (library/macros, docs, other workflows) no longer trigger the full build and QEMU integration tests.

- add explicit paths filters to pull_request and push (main only)
- add workflow_dispatch for manual integration test runs
- cache the EDK2 build; rebuild only when the cache misses

Fix: #224

### DIFF
--- a/.github/workflows/arceboot.yml
+++ b/.github/workflows/arceboot.yml
@@ -2,10 +2,19 @@ name: ArceBoot
 
 on:
   pull_request:
-  push:
     paths-ignore:
       - '**.md'
       - 'LICENSE'
+      - 'library/**'
+      - 'prototyper/**'
+  push:
+    branches: ["main"]
+    paths-ignore:
+      - '**.md'
+      - 'LICENSE'
+      - 'library/**'
+      - 'prototyper/**'
+  workflow_dispatch: {}
 
 jobs:
   build-and-clippy:
@@ -55,8 +64,8 @@ jobs:
       - name: Install required target
         run: rustup target add riscv64gc-unknown-none-elf
 
-      - name: Build ArceBoot with Prototyper payload
-        run: cargo xtask arceboot --payload
+      - name: Build ArceBoot
+        run: cargo xtask arceboot
 
       - name: Set up Python
         uses: actions/setup-python@v5
@@ -68,7 +77,18 @@ jobs:
           sudo apt update
           sudo apt install -y uuid-dev qemu-system-misc
 
+      # Change the version after edk change
+      - name: Cache EDK2 build
+        uses: actions/cache@v4
+        id: cache-edk2
+        with:
+          path: arceboot/edk2/Build
+          key: edk2-${{ hashFiles('arceboot/scripts/test/build_edk2.sh') }}-v1
+          restore-keys: |
+            edk2-
+
       - name: Build EDK2
+        if: steps.cache-edk2.outputs.cache-hit != 'true'
         run: bash arceboot/scripts/test/build_edk2.sh
 
       - name: Generate disk image
@@ -85,7 +105,8 @@ jobs:
         run: |
           timeout 60 qemu-system-riscv64 \
             -m 128M -serial mon:stdio -nographic -machine virt \
-            -bios target/riscv64gc-unknown-none-elf/release/rustsbi-prototyper-payload.elf \
+            -bios default \
+            -kernel target/riscv64gc-unknown-none-elf/release/arceboot.bin \
             -device virtio-blk-pci,drive=disk0 \
             -drive id=disk0,if=none,format=raw,file=arceboot/disk.img \
             > arceboot/qemu-hello-riscv.log 2>&1 || true
@@ -110,7 +131,8 @@ jobs:
         run: |
           timeout 60 qemu-system-riscv64 \
             -m 128M -serial mon:stdio -nographic -machine virt \
-            -bios target/riscv64gc-unknown-none-elf/release/rustsbi-prototyper-payload.elf \
+            -bios default \
+            -kernel target/riscv64gc-unknown-none-elf/release/arceboot.bin \
             -device virtio-blk-pci,drive=disk0 \
             -drive id=disk0,if=none,format=raw,file=arceboot/disk.img \
             > arceboot/qemu-allocate.log 2>&1 || true
@@ -135,7 +157,8 @@ jobs:
         run: |
           timeout 60 qemu-system-riscv64 \
             -m 128M -serial mon:stdio -nographic -machine virt \
-            -bios target/riscv64gc-unknown-none-elf/release/rustsbi-prototyper-payload.elf \
+            -bios default \
+            -kernel target/riscv64gc-unknown-none-elf/release/arceboot.bin \
             -device virtio-blk-pci,drive=disk0 \
             -drive id=disk0,if=none,format=raw,file=arceboot/disk.img \
             > arceboot/qemu-hello.log 2>&1 || true


jack-wang-176:
Well here I give some reason for my code. I am not sure it is appropriate or not that write this in pr descption above. And I just  delete the 'click for commit rules checklist' ,emm,hope do right. So for my code, I use way already be done in openeuler-aia.yml and add cache for edk2. The key question I want to get confirmation is that the way of building cache-key. Now  build_edk2.sh change will change the key but the if edk2 change we need to update key version. It is not elegent certainly but indeed a minimum change. Another way I consider is pull all edk2 commit history but I am not sure it is a good idea or not. Anyway thanks for your help and time, looking forward to your help and more elegant way to build cache-key is appreciate

.github/workflows/arceboot.yml
```diff
@@ -68,7 +89,18 @@ jobs:
...
+        id: cache-edk2
+        with:
+          path: arceboot/edk2/Build
+          key: edk2-${{ hashFiles('arceboot/scripts/test/build_edk2.sh') }}-v1
+          restore-keys: |
+            edk2-
+
       - name: Build EDK2
+        if: steps.cache-edk2.outputs.cache-hit != 'true'
         run: bash arceboot/scripts/test/build_edk2.sh
```
luojia65, line 92: Make minimum modifications, if you want to change this part, create another pull request instead.

.github/workflows/arceboot.yml
```diff
@@ -2,10 +2,31 @@ name: ArceBoot
...
+      - 'xtask/**'
+      - 'prototyper/**'
+      - 'library/rustsbi/**'
+      - 'library/sbi-rt/**'
+      - 'library/sbi-spec/**'
+      - 'Cargo.toml'
+      - 'Cargo.lock'
+      - 'rust-toolchain.toml'
+      - '.github/workflows/arceboot.yml'
+  workflow_dispatch: {}
```
luojia65: That didn't change what path should arceboot CI follow, you just made a list of all existing files, sometimes when the project structure changes, forgetting to update workflow files would result in incorrect behaviors.
I suggest just to accept arceboot/** only OR ignore library/**, prototyper/**.

luojia65 (changes requested):

.github/workflows/arceboot.yml
jack-wang-176: already changed ignoring library and prototyper

.github/workflows/arceboot.yml
jack-wang-176: This change accoring to issue request, use cache for edk2 to let workflow run quickly. The pr is aim to #224 issue so I put this here.

luojia65 (approved):
LGTM

.github/workflows/arceboot.yml
```diff
@@ -2,10 +2,15 @@ name: ArceBoot
...
 on:
   pull_request:
+    paths-ignore:
+      - 'library/**'
+      - 'prototyper/**'
   push:
+    branches: ["main"]
     paths-ignore:
-      - '**.md'
-      - 'LICENSE'
```
Plucky923, line 8: Why was the ignore in the original document deleted?

Plucky923:
One concern: the workflow now ignores `prototyper/**` and `library/**`, but the integration job still builds `rustsbi-prototyper-payload.elf` from the current repository HEAD and uses it for all QEMU tests. Therefore, these paths remain actual test inputs.

I suggest decoupling ArceBoot CI from the in-tree prototyper first, for example by using a pinned SBI implementation. A separate prototyper–ArceBoot integration workflow can be added later if needed. Until then, the path filters do not match the workflow's actual dependencies.

.github/workflows/arceboot.yml
luojia65: @jack-wang-176: restore **.md and LICENSE files, don't ignore them

jack-wang-176: Already restored the original `**.md` and `LICENSE` path-ignore entries.I also decoupled the integration tests from the in-tree Prototyper before keeping `library/**` and `prototyper/**` ignored. The WSL2/QEMU verification passed for HelloRiscv, AllocatePage, and Hello.

jack-wang-176:
> One concern: the workflow now ignores `prototyper/**` and `library/**`, but the integration job still builds `rustsbi-prototyper-payload.elf` from the current repository HEAD and uses it for all QEMU tests. Therefore, these paths remain actual test inputs.
> 
> I suggest decoupling ArceBoot CI from the in-tree prototyper first, for example by using a pinned SBI implementation. A separate prototyper–ArceBoot integration workflow can be added later if needed. Until then, the path filters do not match the workflow's actual dependencies.

Thanks for pointing this out. You were right that the previous path filters did not match the workflow's actual dependencies.I updated the workflow to decouple the ArceBoot integration tests from the in-tree RustSBI Prototyper.The integration job now runs `cargo xtask arceboot` without `--payload` and QEMU uses its built-in OpenSBI with the generated `arceboot.bin` as the kernel.

I also verified the complete runtime path in WSL2 with QEMU 10.2.1. All three EDK2 integration cases passed.

Plucky923 (approved):
LGTM!

.github/workflows/arceboot.yml
```diff
@@ -110,7 +131,8 @@ jobs:
         run: |
           timeout 60 qemu-system-riscv64 \
             -m 128M -serial mon:stdio -nographic -machine virt \
-            -bios target/riscv64gc-unknown-none-elf/release/rustsbi-prototyper-payload.elf \
+            -bios default \
```
luojia65, line 134: In the future we may introduce running ArceBoot using a variety of BIOS'es. This is good for now.

Edit: https://github.com/rustsbi/rustsbi/issues/228

luojia65 (approved):
LGTM. I left a comment for future works.

Plucky923:
The current CI seems to be blocked, possibly due to network issues? Could you trigger the CI again? Once the CI passes, we can merge.

luojia65:
@Plucky923 I triggered CI again just now. Let's see what happens. I suspect this is caused by the GitHub outage incident yesterday.

luojia65:
Now that CI passes, this pull request can be merged.

PR #226, opened 2026-08-17
feat(aia): add indirect CSR accessors for M, S and VS modes

luojia65:
Move IMSIC indirect registers into csrind; add M-, S- and VS-mode accessors.


library/riscv-aia/src/csrind.rs
```diff
@@ -0,0 +1,96 @@
...
+/// S-mode accessors for the supervisor-level interrupt file.
+mod supervisor {
+    use crate::register::{
+        sireg::{self, Sireg},
+        siselect::{self, Siselect},
+    };
+    use riscv::interrupt::supervisor as interrupt;
+
+    #[inline]
+    pub(super) unsafe fn read_ind(reg_id: usize) -> usize {
```
Plucky923, line 71: Please add `# Safety` documentation to these `unsafe fn`s (here and the same in `supervisor`/`guest`). Callers need to know the preconditions: the hart must implement the corresponding AIA extension (Smaia for machine, Ssaia for supervisor, and the hypervisor extension with a guest interrupt file for VS-mode); otherwise selecting a reserved `*iselect` value and accessing `*ireg` raises an illegal-instruction exception.

Plucky923:
This PR removes two riscv-aia APIs that are still used by the in-tree rustsbi-prototyper:

- `peripheral::imsic::system::AddressLayout`
- `peripheral::imsic::msi::encode_le`

Since Prototyper depends on riscv-aia by path, it no longer compiles. The failures in the Prototyper, openEuler AIA, and ArceBoot integration workflows all originate from this build failure.

Please keep the in-tree consumer and the library change consistent in this PR: either update Prototyper to stop using these APIs, or retain them until the migration is done.

The migration should be straightforward:

Move the IMSIC address-layout data and address calculation into Prototyper, where the values are obtained from the device tree and used.
Replace msi::encode_le(iid) with u32::from(iid).

The repository should remain buildable after removing these APIs.

library/riscv-aia/src/csrind/eie.rs
```diff
@@ -0,0 +1,44 @@
+//! External interrupt-enable registers.
+
+/// Indirect CSR base identifier for `eie[n]` registers (eie0 to eie63).
+pub const EIE_BASE: usize = 0xC0;
```
Plucky923, line 4: For `eidelivery` and `eithreshold`, this PR generates ready-made `read`/`write` functions for M/S/VS mode. But for `eie` and `eip` it generates nothing — only the value types and the `EIE_BASE`/`EIP_BASE` constants.

So enabling an interrupt identity — the most common thing a driver does with an IMSIC — has no support from the crate. The user has to write the whole indirect-access sequence by hand:

```rust
// enable interrupt identity 5 in S-mode: what a user must write today
riscv::interrupt::supervisor::free(|| unsafe {
    let prev = siselect::read();
    siselect::write(Siselect::from_bits(0xC0)); // select eie[0]
    let v = sireg::read().bits();
    sireg::write(Sireg::from_bits(v | (1 << 5))); // enable identity 5
    siselect::write(prev);
});
```

Could the same generated `read`/`write` be added for `eie`/`eip` too? They just need an index parameter, since there are up to 64 of them: `read(index)` would read `EIE_BASE + index`. If you prefer to do this in a follow-up PR, that's fine — just mention the plan in the module docs so users know it's coming.

library/riscv-aia/src/csrind/eie.rs
```diff
@@ -0,0 +1,44 @@
...
+    /// Returns whether the interrupt with the given index (0-31) is enabled.
+    pub const fn is_enabled(self, index: u32) -> bool {
+        if index >= 32 {
+            return false;
+        }
+        (self.0 & (1 << index)) != 0
+    }
+
+    /// Sets the enable status for the interrupt with the given index (0-31).
+    pub const fn set_enabled(mut self, index: u32, enabled: bool) -> Self {
```
Plucky923, line 36: The four new `csrind` value-type files ship no tests, while nearly every other register type in this crate has them (`miselect.rs`, `iid.rs`) — and this PR itself tests the `Imsic` layout, so this looks like an oversight.

Most worth pinning is the out-of-range behavior:

```rust
// index >= 32 is silently ignored — pin that
assert_eq!(Eie::from_bits(0).set_enabled(32, true).bits(), 0);
assert!(Eie::from_bits(0).set_enabled(31, true).is_enabled(31));
```

Same for `eip.rs`, and for the `Eidelivery`/`Eithreshold` predicates. Note `Library.yml` doesn't run `cargo test -p riscv-aia` at all — worth a follow-up job.

library/riscv-aia/src/csrind/eithreshold.rs
```diff
@@ -0,0 +1,73 @@
...
+        self.0
+    }
+
+    /// Returns the threshold value.
+    pub const fn threshold(self) -> usize {
+        self.0
+    }
+}
+
+macro_rules! impl_eithreshold_accessors {
```
Plucky923: `impl_eithreshold_accessors!` is a byte-for-byte copy of `impl_eidelivery_accessors!` (eidelivery.rs:52) — the two differ only in the three names (`EITHRESHOLD`/`Eithreshold` vs `EIDELIVERY`/`Eidelivery`) and two doc strings.

Since any future change to the accessor pattern (e.g. the indexed variant for `eip`/`eie`) would have to be applied to each copy, would you consider hoisting one parameterized macro into `csrind.rs` instead?

```rust
macro_rules! impl_ind_accessors {
    ($select:ident, $ty:ident, $name:literal, $($mode:ident => ($doc:literal, $safety:literal)),+) => { /* same body */ }
}
```

library/riscv-aia/src/peripheral/imsic.rs
```diff
@@ -10,508 +10,67 @@ use volatile_register::WO;
...
     /// IMSIC address layout parameters.
-    ///
-    /// IMSIC interrupt files are arranged in memory according to specific patterns
-    /// to allow efficient address calculation.
     #[derive(Clone, Copy, Debug, PartialEq, Eq)]
     pub struct AddressLayout {
         /// Base address for machine-level interrupt files.
         pub machine_base: usize,
         /// Base address for supervisor-level interrupt files.
         pub supervisor_base: usize,
```
Plucky923: `supervisor_base`, `guest_base` and `guest_offset_bits` no longer have any readers: their consumers (`supervisor_interrupt_file_address`, `guest_interrupt_file_address`) were removed by this PR, and nothing else in the crate touches them — the only surviving method, `machine_interrupt_file_address`, reads just `machine_base`, `group_bits` and `hart_offset_bits`.

But struct fields are required at construction, so every user must still fill them with dummy values — the prototyper already does (`platform/mod.rs:608`: `supervisor_base: 0, guest_base: None, guest_offset_bits: 0`).

Since the restore was minimal for the prototyper's sake, would you also drop these three fields here and at the construction site? (In-tree impact: delete the three dummy assignments; no logic reads them.)

Plucky923 (approved):
LGTM!

Issue #219, opened 2026-08-04
Refactor RustSBI Prototyper subproject


luojia65:
RustSBI Prototyper originally used a monolithic design that combined boot, platform initialization, device drivers, CSR and MMIO access, trap handling, and SBI policy within a single M-mode program.

Following the [design document](https://www.plucky.wang/projects/firmware/rethink-how-to-construct-security-firmware-by-rust/) and [PR #215](https://github.com/rustsbi/rustsbi/pull/215), this issue tracks splitting Prototyper into a runtime for security-sensitive mechanisms, a safe firmware policy layer for platform composition and SBI services, device drivers, and unified build and testing tools.

The originally proposed `rustsbi-machine` layer is being implemented as `rustsbi-firmware-runtime` under `firmware/runtime/`, following the [Runtime naming and layering discussion](https://github.com/rustsbi/rustsbi/issues/219#issuecomment-5461302437). Runtime extraction and Prototyper integration have landed incrementally; the broader refactoring remains in progress. Checked items below refer to merged work, with progress reviewed on 2026-09-18.

- [ ] Refine RustSBI libraries
  - [ ] Refine `riscv-aia` crate
    - [x] Refine `riscv-aia` CSRs in `register` module #220
    - [x] Add per-privilege-mode indirect CSRs #226
    - [x] Add `Mtopei::claim` for atomic IMSIC interrupt claiming #258; integrate atomic claiming during platform-resource refactoring #268
    - [ ] Fix peripheral definitions in `riscv-aia`
  - [x] Implement flags for the `pmu` module in `sbi-rt` #230
  - [ ] Add non-exhaustive `HartState` enum in `sbi-spec`
  - [ ] Improve `sbi-testing`
    - [x] Port `sbi-testing` to RV32I architecture #221
    - [ ] Move `sbi-testing` submodule path
  - [x] Support a configurable RustSBI crate path in the derive macro, with static/dynamic dispatch tests #280
- [ ] Introduce the firmware Runtime layer (originally proposed as `rustsbi-machine`)
  - [x] Separate entry mechanism from policies in `rustsbi-prototyper` #223
  - [x] Rename the `prototyper/` directory group to `firmware/` ahead of the crate split #250
  - [x] Introduce the workspace-private `rustsbi-firmware-runtime` crate with validated physical ranges, supervisor-memory access, and typed bounded MMIO handles #266
[...]

PR #230, opened 2026-08-19
feat(pmu)!: integrate spec flags with sbi-rt

luojia65:
Align the PMU flag type names with sbi-rt traits and implement the traits for direct use.

BREAKING CHANGE: use ConfigFlags, StartFlags, and StopFlags for PMU flag types.
Refs: #219


library/sbi-rt/src/pmu.rs
```diff
@@ -358,10 +372,35 @@ pub trait StopFlags {
...
+
+    use super::{ConfigFlags as _, StartFlags as _, StopFlags as _};
+
+    #[test]
+    fn spec_flags_implement_runtime_flag_traits() {
+        let config = ConfigFlags::SKIP_MATCH | ConfigFlags::AUTO_START;
+        let start = StartFlags::INIT_VALUE | StartFlags::INIT_SNAPSHOT;
+        let stop = StopFlags::RESET | StopFlags::TAKE_SNAPSHOT;
+
+        assert_eq!(config.raw(), config.bits());
```
Plucky923: The internal implementation of config.raw() is actually just bits(); this test cannot actually catch any bugs.

Plucky923:
Thanks for this PR. I have one API-naming concern and one test suggestion.

**Avoid the trait/type name collision**

After this rename, the following public items have the same names:

```rust
sbi_rt::ConfigFlags                 // trait
sbi_spec::pmu::flags::ConfigFlags   // bitflags type
```
The same applies to `StartFlags` and `StopFlags`. Because traits and structs share Rust's type namespace, users who import both under their natural names get E0252:
```rust
use sbi_rt::ConfigFlags;
use sbi_spec::pmu::flags::ConfigFlags;
// error[E0252]: the name `ConfigFlags` is defined multiple times
```

Importing the trait as _ works for method resolution, as the test in this PR demonstrates:
```rust
use sbi_rt::ConfigFlags as _;
```
However, this workaround is not particularly discoverable.

I suggest keeping the concise names for the concrete bitflag types that callers construct frequently, and renaming the runtime traits to describe their role as accepted parameter types:
```rust
pub trait ConfigFlagsParam {
    fn raw(&self) -> usize;
}


pub trait StartFlagsParam {
    fn raw(&self) -> usize;
}


pub trait StopFlagsParam {
    fn raw(&self) -> usize;
}
```
This also reads naturally at the main point where the trait name appears:
```
pub fn pmu_counter_start<T: StartFlagsParam>(...)
```


**Test the typed and raw paths against each other**

The current assertions:
```
assert_eq!(config.raw(), config.bits());
```
only verify that the new implementation forwards directly to .bits(). They do not exercise the compatibility this PR introduces: a typed sbi-spec value and a raw integer should be accepted through the same parameter trait and produce the same ABI value. Equivalent checks can be added for StartFlags and StopFlags. Using a helper also avoids ambiguity from calling .raw() directly on usize, since usize implements all three traits when integer-impls is enabled.

**future**

If sbi-spec eventually needs to expose definitions from multiple SBI specification versions side by side, I would prefer versioned module paths:
```rust
flags::v3::StartFlags
```
rather than version suffixes:
```rust
flags::StartFlagsV3
```
The version is part of the namespace rather than the type's semantic name, and the module can later be introduced without breaking the existing path through a re-export such as pub use v3::*;. No change is requested for this PR; this is only a note for future API evolution.

Plucky923 (approved):
LGTM!

PR #235, opened 2026-08-23
refactor(prototyper): decouple board discovery from the SBI extension set

Plucky923:
The prototyper's `Platform` struct fuses three unrelated concerns in
one global: board discovery from the FDT (`info: BoardInfo`), the SBI
extension set (`sbi: SBI`), and the boot-sync flag (`ready`). Every
layer reaches through `static mut PLATFORM` — the sbi layer included,
which has to walk a platform global just to find its own extensions.
The fused `Platform::init` likewise mixes FDT discovery with the
construction of every SBI extension, so neither half can be read or
changed without the other.

This splits them by ownership:

- `platform` keeps board discovery: FDT parsing fills `BOARD_INFO`
  behind `board_info()`, and the boot-sync flag becomes a plain
  `READY: AtomicBool`.
- the sbi layer owns its extension set: `SbiDispatcher` (was `SBI`)
  is constructed from the discovered board info and published once
  via `spin::Once`, with safe fine-grained accessors; this half now
  carries zero unsafe.
- `platform/boot.rs::init_board` sequences the phases (discovery and
  construction interleave as before, e.g. the console and logger come
  up before the discovery steps whose panics need printing).

Consumer sites are migrated off `PLATFORM` accordingly, and the CI
boot gate learns to assert the dispatcher-backed extension lines so a
regression in the new construction path fails loudly instead of
booting green with silently absent extensions.

Wire ABI, init ordering, and the boot log are unchanged — verified
with the full local QEMU matrix (payload/dynamic/jump × test/bench)
and a byte-identical boot log on `aia=aplic-imsic`.

`BOARD_INFO` intentionally stays `static mut` for now (confined to
its accessors); hardening it is the planned next step.

<details>

</details>

### DIFF
--- a/.github/scripts/prototyper-qemu-boot.sh
+++ b/.github/scripts/prototyper-qemu-boot.sh
@@ -71,6 +71,14 @@ run_once() {
   grep -F 'Hello RustSBI!' "$log_file" || return 1
   grep -F "Platform HART Count           : $smp" "$log_file" || return 1
 
+  # Dispatcher-backed extension wiring: these lines render from the
+  # published SBI_DISPATCHER (presence chains + Once publish). A missing
+  # line means an extension was silently dropped during boot assembly.
+  grep -F 'Platform HSM Extension        : Available' "$log_file" || return 1
+  grep -F 'Platform RFence Extension     : Available' "$log_file" || return 1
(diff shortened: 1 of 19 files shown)


luojia65 (approved):
LGTM, we can discuss on further coding convention in future contributions.

PR #236, opened 2026-08-24
refactor(prototyper): publish board facts via Once, split runtime CPU table

Plucky923:
`BOARD_INFO` was the last user-code `static mut` from #235, with one
documented race: `refresh_cpu_features` rewrites `cpu_enabled` after
`READY`, so concurrent readers could see a torn table.

Six of seven `BoardInfo` fields are freeze-after-discovery facts;
only `cpu_enabled` is rewritten. The containers now match:

- `BOARD_INFO` becomes `spin::Once<BoardInfo>`, published in
  `init_board` before the `READY` release (same pattern as
  `SBI_DISPATCHER`); `board_info()` still returns `&'static
  BoardInfo`, so fact readers are unchanged.
- `cpu_enabled` becomes `CPU_ENABLED: RwLock<Option<_>>`; the
  post-`READY` refresh rewrites it under the write lock, so readers
  take atomic snapshots — the race is closed.

The platform module now has zero `static mut` and zero `unsafe`, and
the crate-level `#![allow(static_mut_refs)]` is narrowed to the four
modules that still touch the trap-stack array and heap pool.

Verified with the full QEMU matrix; boot logs byte-identical.

<details>

</details>

### DIFF
--- a/prototyper/prototyper/src/main.rs
+++ b/prototyper/prototyper/src/main.rs
@@ -2,7 +2,6 @@
 #![feature(fn_align)]
 #![no_std]
 #![no_main]
-#![allow(static_mut_refs)]
 
 extern crate alloc;
 #[macro_use]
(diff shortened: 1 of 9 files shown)


luojia65 (approved):
LGTM

PR #240, opened 2026-08-28
feat(rpmi): add RISC-V Platform Management Interface crate

13584452567:
Implements the RPMI message protocol (header, message types, error codes, service-group identifiers), the shared-memory queue (SMQ) transport with A2P/P2A rings, a mailbox controller for normal/posted request patterns, and the Base service-group client (probe/get spec version/get platform info). The reference implementation mirrored is OpenSBI lib/utils/mailbox (rpmi_mailbox.c, fdt_mailbox_rpmi_shmem.c, rpmi_msgprot.h). Used by the prototyper MPXY/CPPC extensions and the K3 RPMI mailbox backend.

<details>

</details>


luojia65 (approved):

Cargo.toml
```diff
@@ -9,6 +9,7 @@ members = [
     "library/riscv-aia",
     "library/riscv-cove",
     "library/riscv-cove-rt",
+    "library/rpmi",
```
Plucky923, line 12: This adds `library/rpmi` to `members`, but nothing actually runs its tests: `.github/workflows/Library.yml` has `cargo test -p` jobs for sbi-spec, sbi-rt, rustsbi, riscv-aia and penglai (none for rpmi), and the crate isn't in `default-members` either, so even a bare `cargo test` at the repo root skips it.

The 518-line spec-lock suite is the only guard that the crate's wire values match the ratified RPMI 1.0 tables. With no job executing it, the failure mode is: someone fat-fingers a constant -> the tests *would* fail -> nothing runs them -> the PR merges green -> a wrong wire value ships. A wrong value in a definitions crate is the expensive kind of bug: it spreads to every downstream consumer, and interoperability failures from it are painful to localize because each side looks correct to itself.

Would you consider adding a `test-rpmi` job mirroring `test-sbi-spec` (checkout@v4 + nightly toolchain + rust-cache + `cargo test -p rpmi --verbose`), and optionally adding the crate to `default-members` so a bare `cargo test` also covers it?

13584452567: https://github.com/rustsbi/rustsbi/pull/240/commits/f1892b0bf6d99737c1e3a0ab706882f2bceeea83

library/rpmi/src/lib.rs
```diff
@@ -0,0 +1,78 @@
+#![doc = include_str!("../README.md")]
+#![no_std]
+#![deny(missing_docs, unsafe_code, unstable_features)]
```
Plucky923: nit: `unsafe_code` is denied here but still locally overridable via `#[allow(unsafe_code)]`. For a crate whose whole contract is "pure definitions, no unsafe", `#![forbid(unsafe_code)]` would make that invariant non-overridable at zero cost.

@luojia65 while looking at this I checked the other library crates -- none of them declare the `unsafe_code` lint at all. From a quick audit:

- `riscv-cove`, `penglai` and `rustsbi` (its only `unsafe` occurrences are in doc examples, which don't trigger the lint) currently have zero unsafe in real code and could take `#![forbid(unsafe_code)]` as-is;
- `sbi-spec` has two deliberate `pub unsafe fn`s mirroring std's unchecked unwraps, so it would need a module-level approach or a documented exception;
- the rest (`sbi-rt`, `sbi-testing`, `riscv-aia`, `riscv-cove-rt`, `pmpm`, `smm`) use unsafe as part of their job, so crate-level forbid doesn't apply -- module-level forbid in zero-unsafe modules would be the pattern there.

Since this crate would be the first to declare the lint, it would effectively set the precedent for the follow-up pass.

library/rpmi/src/message.rs
```diff
@@ -0,0 +1,378 @@
...
+    pub const fn new(
+        service_group_id: u16,
+        service_id: u8,
+        flags: u8,
+        data_len: u16,
+        token: u16,
+    ) -> Option<Self> {
+        let message_type = (flags & Self::MESSAGE_TYPE_MASK) >> Self::MESSAGE_TYPE_SHIFT;
+        let is_notification = message_type == MessageType::Notification.bits();
+        if !(data_len as usize).is_multiple_of(DATA_LEN_ALIGNMENT)
```
Plucky923: question: `usize::is_multiple_of` was stabilized in Rust 1.87, so this call -- and the same one in `EventHeader::new` at line 303 -- quietly sets the crate's minimum toolchain at 1.87 with no declared `rust-version`. Verified locally: the crate compiles on 1.87.0 and fails on 1.86.0 with four errors -- `E0658 use of unstable library feature unsigned_is_multiple_of` plus const-fn stability errors -- where the compiler's suggested fix (`#![feature(unsigned_is_multiple_of)]`) conflicts with the crate's own `#![deny(unstable_features)]`, and a consumer on 1.85/1.86 hits this inside a dependency they can't patch.

Since edition 2024 alone would allow 1.85: is `rust-version = "1.87"` in the manifest the intent, or is `data_len % 4 == 0` preferable to keep the floor lower?

library/rpmi/src/tests.rs
```diff
@@ -0,0 +1,518 @@
...
+            request_forward::RETRIEVE_CURRENT_MESSAGE,
+            request_forward::COMPLETE_CURRENT_MESSAGE,
+        ],
+        [1, 2, 3]
+    );
+}
+
+#[test]
+fn subtle_v1_encodings_are_locked_down() {
+    assert_eq!(hsm::hart_state::STARTED, sbi_spec::hsm::hart_state::STARTED);
```
Plucky923, line 308: suggestion: the spec-lock suite covers service/group IDs and flag words thoroughly, but several constant families have no locked-value assertions at all: all 22 `cppc::register` IDs (0x00-0x14 plus `TRANSITION_LATENCY = 0x8000_0000`), the `cppc::fast_channel` layout constants (only `DOORBELL_WIDTH_32` is asserted), `system_reset::reset_type` and `system_suspend` values (`SUSPEND_TO_RAM`, `CUSTOM_RESUME_ADDRESS`), the event IDs in `base::event` / `request_forward::event`, and `notification_request_state` / `notification_state`. A typo in any of these would pass CI today.

Three suggestions for this file:
1. Replace the assertion above with the literal values -- it compares the `pub use` re-export with itself and always passes, so it currently locks nothing.
2. Extend the `service_ids_match_ratified_v1`-style coverage to the families listed above.
3. Optional: the test literals are transcribed by the same author as the constants, so they guard against future edits but not initial transcription errors -- embedding an independent reference (e.g. values from OpenSBI's `rpmi_msgprot.h`) would close that gap too.

library/rpmi/src/device_power.rs
```diff
@@ -0,0 +1,90 @@
...
+
+/// Mask of the state value in a power-state word.
+const STATE_VALUE_MASK: u32 = 0x0000_ffff;
+/// Bit indicating that context is lost in the selected state.
+pub const CONTEXT_LOST: u32 = 1 << 16;
+
+/// A device-power state value.
+#[repr(u16)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash)]
+pub enum StateValue {
```
Plucky923, line 31: I checked the spec's power-state encoding table before commenting: values 0x0001-0x0002 are reserved-must-be-zero, so rejecting them in `TryFrom` -- and locking `try_from(1) == Err(1)` in the tests -- is correct as-is.

The gap is at the other end of the range: 0x1000-0xFFFF are vendor-specific states, i.e. spec-valid values a platform may legitimately report through GET_STATE or accept through SET_STATE. Currently they fall into the same `Err(raw)` as reserved values, so a caller can't distinguish a legitimate vendor state from a protocol violation without re-implementing the range check -- and the 0x1000 bound is not defined as a constant anywhere in the crate (its only occurrence is a test literal that itself labels `Err(0x1000)` an `unknown_state`), so callers would have to hard-code it from the spec.

`system_reset::reset_type` already models its own vendor range with `VENDOR_SPECIFIC_START`/`END` constants. Would it make sense to mirror that here -- at minimum `VENDOR_SPECIFIC_START: u16 = 0x1000` / `VENDOR_SPECIFIC_END: u16 = 0xffff`, or a dedicated vendor representation if you prefer the type itself to carry the distinction?

Plucky923 (approved):
LGTM!

PR #248, opened 2026-08-29
refactor(prototyper): move the hart cells into the trap-stack base

Plucky923:
The per-hart cell primitives (`HsmCell`, `RFenceCell`, `Fifo`) are synchronization primitives stored in the per-hart slot — they move from the business files to `trap_stack`, the module that owns the slots. The business files keep only SBI policy; the CSR/fence unsafe they still carry is the next step's scope.

Stacked on #239 (merged).


luojia65 (approved):
LGTM

PR #252, opened 2026-08-29
refactor(prototyper): replace the rfence Fifo with a capped VecDeque

Plucky923:
The rfence FIFO was a standalone generic data type (Fifo<T> with its own error enum and MaybeUninit slots) serving a single caller. Delete the abstraction: RFenceCell now holds a Mutex<VecDeque> bounded by QUEUE_CAP, eliminating the Fifo type, FifoError, and the unsafe assume_init_read. The full-queue retry (drain via IPI) stays in the rfence layer where it belongs.

### DIFF
--- a/firmware/prototyper/src/sbi/trap_stack.rs
+++ b/firmware/prototyper/src/sbi/trap_stack.rs
@@ -8,9 +8,10 @@ use crate::riscv::current_hartid;
 use crate::sbi::hart_context::{HartContext, HartLocal, NextStage};
 use crate::sbi::rfence::RFenceContext;
 use crate::sbi::trap::fast_handler;
+use alloc::collections::VecDeque;
 use core::cell::UnsafeCell;
 use core::hint::spin_loop;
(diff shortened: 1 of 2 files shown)


firmware/prototyper/src/sbi/trap_stack.rs
```diff
@@ -134,20 +134,18 @@ pub enum FifoError {
 }
 
 /// A fixed-size FIFO (First In First Out) queue implementation.
-pub struct Fifo<T: Copy + Clone> {
-    data: [MaybeUninit<T>; FIFO_SIZE],
+pub struct Fifo<T: Copy> {
+    data: [Option<T>; FIFO_SIZE],
```
luojia65: This takes extra byte on struct, as Option<T> uses one byte for each FIFO element. We should save stack size if possible on bare-metal software, as resources are maybe limited.

luojia65 (approved):

PR #254, opened 2026-08-30
fix(prototyper): implement FWFT feature locking

13584452567:
Track FWFT probe and lock state per hart, validate standard feature values, and preserve idempotent set behavior after locking. Restore controlled menvcfg fields when a hart context is reset.

Also correct the standard FWFT feature ID and value bit assignments required by the SBI 3.0 specification.

<details>

</details>

### DIFF
--- a/firmware/prototyper/src/sbi/fwft.rs
+++ b/firmware/prototyper/src/sbi/fwft.rs
@@ -1,7 +1,7 @@
 use rustsbi::SbiRet;
 use sbi_spec::fwft::feature_type;
 
-use crate::riscv::csr::CSR_MENVCFG;
+use crate::riscv::csr::{self, CSR_MENVCFG};
 use crate::sbi::early_trap::{TrapInfo, csr_read_allow, csr_write_allow};
 use crate::sbi::trap_stack::with_current;
 
@@ -47,6 +47,13 @@ impl FwftState {
     fn lock(&mut self, feature_id: u32) {
         self.locked |= Self::mask(feature_id);
     }
+
+    pub(crate) fn reset(&mut self) {
+        // FWFT locks survive non-retentive suspend/resume. Probe results are
+        // cheap to rebuild and may depend on the resumed hart's CSR state.
+        self.probed = 0;
+        self.supported = 0;
+    }
 }
 
 /// Firmware Features extension backed by `medeleg` and `menvcfg`.
@@ -62,7 +69,7 @@ impl SbiFwft {
             return;
         };
         let controlled = ENVCFG_LPE | ENVCFG_SSE | ENVCFG_PMM | ENVCFG_DTE | ENVCFG_ADUE;
-        let _ = Self::menvcfg_write(current & !controlled);
+        let _ = Self::menvcfg_write(current, current & !controlled);
     }
 
     fn has_s_mode() -> bool {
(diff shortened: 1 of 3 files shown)


firmware/prototyper/src/sbi/hart_context.rs
```diff
@@ -75,13 +78,15 @@ impl HartLocal {
...
+        self.fwft_state = FwftState::new();
     }
 
     #[inline]
     pub fn reset(&mut self) {
         self.ipi_reset();
         self.rfence_reset();
         self.pmu_state_reset();
+        self.fwft_state = FwftState::new();
     }
```
luojia65, line 90: The `reset` function here would also be called in non-retentive resume, resulting in possibly locked FwftState to be unlocked. This does not follow the RISC-V SBI 3.0 standard where FWFT locks should be preserved only before hart reset or initial boot (non-retentive resume is not considered a hart reset).

I'll explain the call chain using existing source code:

```rust
pub fn reset(&mut self) { ... }
// is called by ...
pub fn reset_hart(hart_id: usize) {
    with_hart(hart_id, HartLocal::reset);
}
// is called by ...
impl SbiHsm {
    // non retentive resume
    fn hart_resume(&self, hartid: usize, resume_addr: usize, opaque: usize) -> SbiRet {
        match remote_hsm(hartid) {
            Some(remote) => {
                if remote.resume(NextStage {
                    start_addr: resume_addr,
                    opaque,
                    next_mode: MPP::Supervisor,
                }) {
                    // reset the hart local context to prevent the hart context from being polluted
                    reset_hart(hartid);
                    // boot resume hart from resume addr
                    boot();
                } else {
                    SbiRet::failed()
                }
            }
            None => SbiRet::failed(),
        }
    }
}
```

... where non-retentive resume (no context is preserved) will call `HartLocal::reset` to make locked FWFT bits unlocked by replacing the origin context with a new `FwftState`.

In this case I suggest to implement `fwft_state_reset()` function where we do not create a new FwftState structure to replace the old one, at least we should preserve states of the FWFT locks here.

firmware/prototyper/src/sbi/fwft.rs
```diff
@@ -24,6 +57,14 @@ const ENVCFG_PMM: usize = 0b11 << ENVCFG_PMM_SHIFT;
 pub(crate) struct SbiFwft;
 
 impl SbiFwft {
+    pub(crate) fn reset_current() {
+        let Some(current) = Self::menvcfg_read() else {
+            return;
+        };
+        let controlled = ENVCFG_LPE | ENVCFG_SSE | ENVCFG_PMM | ENVCFG_DTE | ENVCFG_ADUE;
+        let _ = Self::menvcfg_write(current & !controlled);
```
luojia65: After changing `menvcfg.ADUE` the RISC-V standard requires to execute `sfence.vma x0, x0` to synchronize S-mode states, and `hfence.gvma x0, x0` for G- and VS-stages if H extension exists. 

Ref: Section 2.1.1.18. Machine Environment Configuration (menvcfg) Register, The RISC-V Instruction Set Manual, Volume II:
> After changing menvcfg.ADUE, executing an SFENCE.VMA instruction with rs1=x0 and rs2=x0 suffices to synchronize address-translation caches with respect to the altered interpretation of page-table entries' A/D bits. See [Memory-Management Fences](https://docs.riscv.org/reference/isa/v20260120/priv/hypervisor.html#hyp-mm-fences) for additional synchronization requirements when the hypervisor extension is implemented.

And Section 14.1.5.3. Memory-Management Fences in the same document:
> Similarly, if the setting of the PBMTE or ADUE bits in menvcfg are changed, an HFENCE.GVMA instruction with rs1=x0 and rs2=x0 suffices to synchronize with respect to the altered interpretation of G-stage and VS-stage PTEs' PBMT and A/D bit fields, respectively.

luojia65 (changes requested):
Please fix those issues which does not follow the RISC-V SBI 3.0 standard.

luojia65 (approved):

PR #255, opened 2026-08-30
fix(prototyper): validate STA shared memory addresses

13584452567:
Return the SBI-defined errors for misaligned and unrepresentable shared memory addresses. Track per-hart STA shared-memory state so each hart registers its own reporting region. Remove the unused reporting helper until a scheduler can provide real steal-time accounting data.

<details>

</details>

### DIFF
--- a/firmware/prototyper/src/sbi/sta.rs
+++ b/firmware/prototyper/src/sbi/sta.rs
@@ -6,6 +6,17 @@ use sbi_spec::binary::SharedPtr;
 use crate::cfg::NUM_HART_MAX;
 use crate::riscv::current_hartid;
 
+#[repr(C)]
+struct StaShmemData {
+    sequence: u32,
+    flags: u32,
+    steal: u64,
+    preempted: u8,
+    pad: [u8; 47],
+}
+
+const _: () = assert!(core::mem::size_of::<StaShmemData>() == 64);
+
 struct StaShmem {
     lo: AtomicUsize,
     hi: AtomicUsize,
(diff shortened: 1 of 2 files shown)


firmware/prototyper/src/sbi/sta.rs
```diff
@@ -15,13 +23,16 @@ impl rustsbi::Sta for SbiSta {
 
         // All-ones shared pointer disables steal-time reporting.
         if hi == usize::MAX && lo == usize::MAX {
+            STA_SHMEM[current_hartid()].store(0, Ordering::Release);
```
luojia65: Do not use `0` to represent that SHMEM is disabled. If physical address `0x0` can be written, the address `0` should be considered valid. RISC-V SBI standard only defined that if address `(usize::MAX, usize::MAX)` is given, the SHMEM should be considered disabled. Here we should use *one* atomic lock of two `usize`'s for SHMEM address (both the `hi` and `lo` addresses is maybe used in RV32), use both usize::MAX as 'disabled'.

luojia65 (changes requested):
Please fix this issue, other parts LGTM.

firmware/prototyper/src/sbi/sta.rs
```diff
@@ -1,6 +1,53 @@
...
+impl StaShmem {
+    const DISABLED: Self = Self {
+        lo: AtomicUsize::new(usize::MAX),
+        hi: AtomicUsize::new(usize::MAX),
+    };
+
+    fn store(&self, lo: usize, hi: usize) {
+        self.lo.store(lo, Ordering::Release);
+        self.hi.store(hi, Ordering::Release);
+    }
```
luojia65: We can't assume that both `lo` and `hi` are modified atomically. It is possible for two threads that one modifies `lo` while other modifies `hi`. Add a lock for both the `usize`'s.

firmware/prototyper/src/sbi/sta.rs
```diff
@@ -1,6 +1,53 @@
...
+// parts are retained for the SBI-defined disabled value and future wider
+// address support, but only a zero high part can be dereferenced here.
+static STA_SHMEM: [StaShmem; NUM_HART_MAX] = [const { StaShmem::DISABLED }; NUM_HART_MAX];
+
+unsafe fn zero_shmem(addr: usize) {
+    let ptr = addr as *mut u8;
+    for offset in 0..core::mem::size_of::<StaShmemData>() {
+        // SAFETY: the caller validated the complete shared-memory range.
+        unsafe { ptr.add(offset).write_volatile(0) };
+    }
```
luojia65: Manually writing `u8`'s are slow. Use `slice::fill` to include optimizations.

luojia65 (changes requested):

PR #257, opened 2026-08-30
feat(prototyper): add SpacemiT K3 SoC support

13584452567:
Add the K3 board configuration, warm-entry path, privileged register helpers, and low-level SoC initialization required to boot the platform.

<details>

</details>

### DIFF
--- a/firmware/macros/src/lib.rs
+++ b/firmware/macros/src/lib.rs
@@ -92,20 +92,6 @@ fn expand(attribute: TokenStream, item: TokenStream) -> TokenStream {
                 )
             }
 
-            /// SpacemiT K3 warm-boot entry, referenced by the cluster
-            /// RVBADDR registers after platform initialization.
-            #[doc(hidden)]
-            #[unsafe(naked)]
-            #[unsafe(link_section = ".text.warmboot")]
-            #[unsafe(export_name = "_start_warm_k3")]
-            unsafe extern "C" fn __rustsbi_prototyper_start_warm_k3() -> ! {
-                ::core::arch::naked_asm!(
-                    include_str!("entry/spacemit_k3.S"),
-                    locate_stack = sym crate::sbi::trap_stack::locate,
-                    main = sym __rustsbi_prototyper_main,
-                    hart_boot = sym crate::sbi::trap::boot::boot,
-                )
-            }
         };
     }
 }
@@ -137,14 +123,8 @@ mod tests {
     fn includes_startup_assembly() {
         let expanded = normalized(expand(TokenStream::new(), POLICY_FN.parse().unwrap()));
         assert_eq!(expanded.matches("export_name=\"_start\"").count(), 1);
-        assert_eq!(
-            expanded.matches("export_name=\"_start_warm_k3\"").count(),
-            1
-        );
         assert!(expanded.contains("link_section=\".text.entry\""));
-        assert!(expanded.contains("link_section=\".text.warmboot\""));
         assert!(expanded.contains("include_str!(\"entry/start.S\")"));
-        assert!(expanded.contains("include_str!(\"entry/spacemit_k3.S\")"));
         assert!(expanded.contains("include_str!(\"entry/relocation.S\")"));
     }
 
--- a/firmware/macros/src/lib.rs
+++ b/firmware/macros/src/lib.rs
@@ -91,6 +91,21 @@ fn expand(attribute: TokenStream, item: TokenStream) -> TokenStream {
                     hart_boot = sym crate::sbi::trap::boot::boot,
                 )
             }
+
+            /// SpacemiT K3 warm-boot entry, referenced by the cluster
+            /// RVBADDR registers after platform initialization.
+            #[doc(hidden)]
+            #[unsafe(naked)]
+            #[unsafe(link_section = ".text.warmboot")]
+            #[unsafe(export_name = "_start_warm_k3")]
+            unsafe extern "C" fn __rustsbi_prototyper_start_warm_k3() -> ! {
+                ::core::arch::naked_asm!(
+                    include_str!("entry/spacemit_k3.S"),
+                    locate_stack = sym crate::sbi::trap_stack::locate,
+                    main = sym __rustsbi_prototyper_main,
+                    hart_boot = sym crate::sbi::trap::boot::boot,
+                )
+            }
         };
     }
 }
@@ -122,7 +137,14 @@ mod tests {
     fn includes_startup_assembly() {
         let expanded = normalized(expand(TokenStream::new(), POLICY_FN.parse().unwrap()));
         assert_eq!(expanded.matches("export_name=\"_start\"").count(), 1);
+        assert_eq!(
+            expanded.matches("export_name=\"_start_warm_k3\"").count(),
+            1
+        );
+        assert!(expanded.contains("link_section=\".text.entry\""));
+        assert!(expanded.contains("link_section=\".text.warmboot\""));
         assert!(expanded.contains("include_str!(\"entry/start.S\")"));
+        assert!(expanded.contains("include_str!(\"entry/spacemit_k3.S\")"));
         assert!(expanded.contains("include_str!(\"entry/relocation.S\")"));
     }
 
--- a/firmware/prototyper/src/entry/start.S
+++ b/firmware/prototyper/src/entry/start.S
@@ -4,11 +4,29 @@
 //
 // The 6:/7: signal words are read before BSS is cleared, so they must live
 // in .text.entry (loaded content), not in BSS.
+//
+// LINK ORDER MATTERS: the cold boot path must occupy the head of the image
+// because the FIT entry point equals the image base (0x100000000). The K3
+// warm-boot entry `_start_warm_k3` is placed AFTER the cold path below, so a
+// PMU-woken secondary hart lands on the warm sequence while the boot hart
+// always starts at the cold ceremony.
 
 .option arch, +a
 
 // 1. Mask all interrupts.
 csrw    mie, zero
+
+// Flush local TLB and I-cache: after a warm restart the RAM image and this
+// hart's caches may hold stale state from the previous boot. SFENCE.VMA
+// exists only on harts implementing S, so gate it on misa.S rather than
+// assuming the extension; FENCE.I (Zifencei) is required by all image
+// targets.
+        csrr    t0, misa
+        srli    t0, t0, 18      // 'S' - 'A'
+        andi    t0, t0, 1
+        beqz    t0, 8f
+        sfence.vma
+8:      fence.i
 
 // 2. Elect a single boot hart: the first hart to arrive wins the race
 // (amoadd returns the previous value; nonzero means another hart won).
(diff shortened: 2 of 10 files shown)


firmware/prototyper/src/riscv/spacemit_k3.rs
```diff
@@ -0,0 +1,1520 @@
...
+
+    /// Write the register.
+    #[inline]
+    fn write(&self, val: u32) {
+        // Safety: `self` points at MMIO registers; writes are volatile so
+        // the compiler cannot elide or reorder them.
+        unsafe { core::ptr::addr_of!(self.0).cast_mut().write_volatile(val) }
+    }
+}
+
```
luojia65: Use `UnsafeCell<u32>` or `volatile_register::{RW, RO, WO}` in this case.

firmware/prototyper/src/entry/start.S
```diff
@@ -52,3 +70,62 @@ csrw    mie, zero
   .word    0
 7: // bss ready signal word.
   .word    0
+
+// PMU-woken harts fetch this entry from their cluster RVBADDR. Data cache
+// and snoop must be enabled before AMOs or Rust code execute. BSS and
+// boot-hart election are already complete.
+    .balign  4
+    .globl _start_warm_k3
+_start_warm_k3:
```
luojia65: This function should be in another `.S` source code file. Fix the `_start` symbol at the beginning of whole firmware by using special `.text.entry` section.

firmware/prototyper/src/riscv/spacemit_k3.rs
```diff
@@ -0,0 +1,1520 @@
...
+// ---------------------------------------------------------------------------
+
+/// Machine L2 cache setup register.
+const CSR_ML2SETUP: u16 = 0x7f0;
+/// Machine L2 cache hint register.
+const CSR_ML2HINT: u16 = 0x7f7;
+/// Machine performance control register (vector load cache level).
+const CSR_PERF_CTRL: u16 = 0x7d0;
+/// Machine prefetch control register (L2 prefetch distance).
+const CSR_PREFETCH_CTRL: u16 = 0x7d1;
```
luojia65, line 27: Those could be SpacemiT custom registers. A separate `spacemit-riscv` crate (like `xuantie-riscv`) should be created for those purposes, hosting all of custom CSRs and (either implied or explicitly documented) fields and values.

luojia65 (approved):
This pull request mostly includes undocumented black magic but works for SpacemiT K3 chips. I suggest to separate K3 magic from original bootloading assembly in `start.S`.

We should furtherly introduce a new `spacemit-riscv` crate in future contributions to wrap and explain those CPU specific features (CSRs, etc.) for extensibility of the whole firmware.

Please add screenshots of bootloading on real hardware as a reference. Other parts LGTM.

13584452567:
29b3cba2f1f0205e47584706185e2ccb
5bbb92699d7d8ae19a10ba0adc9d8ec4

CPU specific features (CSRs, etc.) and the black magic codes will be collected together after the spacemit-riscv crate being created. Working on other comments.

firmware/prototyper/src/riscv/spacemit_k3.rs
Plucky923: see #258, we will add new mmio abstraction.

The mmio abstraction cannot be added to the k3 file.

firmware/prototyper/src/riscv/spacemit_k3.rs
```diff
@@ -0,0 +1,1503 @@
...
+/// emulated for S-mode -?return `None` so the fault is redirected back to
+/// S-mode as a real access error (spacemit_k3.c `m_only_ranges[]`, L304-326).
+struct AddrRange {
+    base: usize,
+    size: usize,
+}
+
+const fn addr_range(base: usize, size: usize) -> AddrRange {
+    AddrRange { base, size }
+}
```
Plucky923: I don't understand the necessity of this abstraction. It's clearly a generic abstraction, yet the comment says it was introduced because of spacemit_k3.

firmware/prototyper/src/riscv/spacemit_k3.rs
13584452567: The mmio abstraction has been replaced by volatile_register in https://github.com/rustsbi/rustsbi/pull/257/commits/8bad105e9c5608145c0cd7d5e576b42ed512c300

firmware/prototyper/src/riscv/spacemit_k3.rs
13584452567: You are right that AddrRange itself is generic. It is only a small local representation used by the K3-specific M_ONLY_RANGES table, which mirrors the discontiguous m_only_ranges[] table from the K3 implementation. It was not intended to be a reusable platform abstraction. I will clarify the comment so that the K3-specific part is clearly identified as the range table and the emulation policy, not AddrRange itself.

firmware/prototyper/src/riscv/spacemit_k3.rs
Plucky923: I don't think we should add a third-party crate this easily to a low-level project like RustSBI. Even if it comes from `rust-embedded`, every new dependency adds long-term maintenance and trust costs.

Also, `volatile-register` has a long-standing open soundness issue around reference-based MMIO (`rust-embedded/volatile-register#10`).

Since #258 is already designing an MMIO abstraction, I think K3's requirements should be discussed there. If the current design cannot support K3, we should improve that abstraction together instead of introducing another solution here.

firmware/prototyper/src/riscv/spacemit_k3.rs
Plucky923: I don't think we need a new `AddrRange` type here. It is just another representation of a memory address range and does not provide additional semantics or safety.

A simple `(base, size)` or `Range<usize>` should be enough. If we do need a common address-range abstraction, it should probably be defined in a shared module rather than introduced specifically for K3.

firmware/macros/src/lib.rs
```diff
@@ -91,6 +91,21 @@ fn expand(attribute: TokenStream, item: TokenStream) -> TokenStream {
...
+            #[unsafe(export_name = "_start_warm_k3")]
+            unsafe extern "C" fn __rustsbi_prototyper_start_warm_k3() -> ! {
+                ::core::arch::naked_asm!(
+                    include_str!("entry/spacemit_k3.S"),
+                    locate_stack = sym crate::sbi::trap_stack::locate,
+                    main = sym __rustsbi_prototyper_main,
+                    hart_boot = sym crate::sbi::trap::boot::boot,
+                )
+            }
         };
```
Plucky923, line 95: The entry layer should remain generic and only provide the common firmware entry path.

`_start_warm_k3`, `RVBADDR`, and `spacemit_k3.S` are SoC-specific details and should belong to the K3 platform implementation rather than the common entry layer.

If K3 requires a special warm-boot entry, the common entry layer should expose a generic mechanism for platform-specific entry points instead of directly embedding K3-specific code.

firmware/prototyper/src/riscv/spacemit_k3.rs
Plucky923, line 1: This file is too large and mixes too many responsibilities, including custom CSRs, MMIO registers, power management, cache/coherency setup, warm boot, interrupt state, and platform detection.

These components should be split into smaller modules with clear responsibilities. Keeping all K3-specific logic in one file will make the code harder to review, test, and maintain as more platform features are added.

firmware/prototyper/src/riscv/spacemit_k3.rs
```diff
@@ -0,0 +1,1520 @@
...
+//! `volatile_register::{RW, RO, WO}` rather than raw base-address arithmetic,
+//! keeping access permissions explicit and register layouts testable via
+//! `offset_of!`.
+
+use core::arch::asm;
+use volatile_register::{RO, RW, WO};
+
+// ---------------------------------------------------------------------------
+// Custom CSRs (0x7f0, 0x7f7, 0x7d0, 0x7d1)
+// ---------------------------------------------------------------------------
```
Plucky923, line 18: This style of section separator is not idiomatic Rust.

Rust code should preferably express structure through modules and documentation comments rather than large visual separator comments. If this file needs many separators like this, it is also a sign that the file should probably be split into smaller modules.

I mentioned the comment style issue in a previous PR as well. If this contribution is AI-assisted, please use AI to review the comments before submission, especially for Rust style and unnecessary generated comments.

firmware/prototyper/src/riscv/spacemit_k3.rs
```diff
@@ -0,0 +1,1520 @@
...
+/// Machine performance control register (vector load cache level).
+const CSR_PERF_CTRL: u16 = 0x7d0;
+/// Machine prefetch control register (L2 prefetch distance).
+const CSR_PREFETCH_CTRL: u16 = 0x7d1;
+
+// ML2SETUP bit fields (in addition to the per-hart L2 cache mask)
+const ML2SETUP_IPRF: usize = 1 << 16; // Instruction prefetch enable
+const ML2SETUP_TPRF: usize = 1 << 18; // TLB prefetch enable
+
+// ML2HINT bit fields (core_common.h)
```
Plucky923, line 33: Why is there `(core common.h)` in the comments?

firmware/macros/src/lib.rs
13584452567: I will keep the macro responsible only for the generic _start path and the Rust main bridge, and move the K3 warm-boot entry and assembly inclusion into the K3 platform module.

firmware/prototyper/src/riscv/spacemit_k3.rs
```diff
@@ -0,0 +1,1520 @@
...
+    (0xd4050000usize + 0x10dc) as *const WO<u32>, // CORE7
+    (0xd4050000usize + 0x10e0) as *const WO<u32>, // CORE8
+    (0xd4050000usize + 0x10e4) as *const WO<u32>, // CORE9
+    (0xd4050000usize + 0x10e8) as *const WO<u32>, // CORE10
+    (0xd4050000usize + 0x10ec) as *const WO<u32>, // CORE11
+    (0xd4050000usize + 0x10f0) as *const WO<u32>, // CORE12
+    (0xd4050000usize + 0x10f4) as *const WO<u32>, // CORE13
+    (0xd4050000usize + 0x10f8) as *const WO<u32>, // CORE14
+    (0xd4050000usize + 0x10fc) as *const WO<u32>, // CORE15
+];
```
Plucky923, line 168: There is a lot of repetitive register-layout code here. A small declarative macro could make this much clearer and avoid repeating pointer casts and per-core comments.

For irregular per-hart registers, the macro can describe the offsets as one logical register layout. For regular registers such as `APCR_CORE_VETE_REG`, a simple `base + hart_id * stride` helper is enough.

The macro should describe the hardware layout only; MMIO access itself should still go through the common MMIO abstraction.

firmware/prototyper/src/riscv/spacemit_k3.rs
```diff
@@ -0,0 +1,1539 @@
...
+    const APLIC_SMSICFGADDR: usize = 0x1bc8;
+    const APLIC_SMSICFGADDRH: usize = 0x1bcc;
+
+    #[inline]
+    fn rd(base: usize, off: usize) -> u32 {
+        unsafe { core::ptr::read_volatile((base + off) as *const u32) }
+    }
+
+    // Cold-init sequence for the K3 maplic.
+    // 1. Domain configuration to 0 (interrupt domain disabled).
```
Plucky923, line 1370: Don't add such comments to re-explain the code. If the semantics can be clearly expressed in the code, don't use redundant comments to explain them again. A good piece of code doesn't need comments.

firmware/prototyper/src/riscv/spacemit_k3.rs
```diff
@@ -0,0 +1,1539 @@
...
+    addr_range(0xd4282c00 + 0x3e8, 2 * 4),  // C2_RVBADDR LO/HI
+    addr_range(0xd4282c00 + 0x260, 2 * 4),  // C3_RVBADDR LO/HI
+    addr_range(0xd4282800 + 0xe4, 4),       // PMU_CX_CAPMP_IDLE_CFG1
+    addr_range(0xd4282800 + 0x120, 7 * 4),  // CFG0, IDLE_CFG0/1, WAKEUP0-3
+    addr_range(0xd4282800 + 0x150, 2 * 4),  // CFG2/3
+    addr_range(0xd4282800 + 0x160, 2 * 4),  // IDLE_CFG2/3
+    addr_range(0xd4282800 + 0x20c, 12 * 4), // IDLE_CFG12-15, CX_CFG12-15, WAKEUP12-15
+    addr_range(0xd4282800 + 0x304, 12 * 4), // IDLE_CFG4-7, CX_CFG4-7, WAKEUP4-7
+    addr_range(0xd4282800 + 0x340, 12 * 4), // IDLE_CFG8-11, CX_CFG8-11, WAKEUP8-11
+];
```
Plucky923, line 765: This table contains too many raw addresses and sizes with almost no type or semantic information.

A maintainer should not have to go back to the hardware manual just to understand what `0xd4282800 + 0x20c` or `12 * 4` means. These ranges should be derived from named register definitions or a structured register layout instead of duplicating hardware addresses as magic numbers.

Ideally, the register layout should be the single source of truth, and ranges like `M_ONLY_RANGES` should be derived from it rather than maintained separately.

firmware/prototyper/src/riscv/spacemit_k3.rs
```diff
@@ -0,0 +1,1539 @@
...
+}
+
+/// Clear bits in a CSR (read-modify-write).
+///
+/// `CSR` must be a compile-time constant; see [`csr_write`].
+#[inline]
+unsafe fn csr_clear<const CSR: u16>(bits: usize) {
+    let old = unsafe { csr_read::<CSR>() };
+    unsafe { csr_write::<CSR>(old & !bits) };
+}
```
Plucky923, line 306: Why are these things still needed? They are all universal—aren't they already defined elsewhere?

firmware/prototyper/src/riscv/spacemit_k3.rs
```diff
@@ -0,0 +1,1539 @@
...
+
+/// Parking entry for the first A100 core (hart 8, cluster 2).
+///
+/// The boot hart points `C2_RVBADDR` at this function and asserts
+/// `PMU_CAP_CORE8_WAKEUP`, waking core8 which then performs the vector-load
+/// tuning, votes its cluster/core down, disables caches and snoop, and hangs
+/// in a `wfi` loop.
+///
+/// This function never returns.
+fn boot_entry_dummy() -> ! {
```
Plucky923, line 381: What does "dummy" mean?

firmware/prototyper/src/riscv/spacemit_k3.rs
luojia65: Use UnsafeCell.

Plucky923:
Overall, this change contains many ad-hoc abstractions and helper functions. Much of the code looks like a mechanical translation of the original C implementation rather than a design that takes advantage of Rust.

Rust allows many invariants to be expressed and checked at compile time through types, traits, const evaluation, and macros. For example, in this file the same register information is repeated across typed pointer tables, `M_ONLY_RANGES`, and tests. Register address, access permission, and privilege should ideally be described once, and the corresponding accessors and policies can be generated from that single definition. This would remove a large amount of raw addresses, duplicated metadata, and runtime/helper logic while preventing invalid operations at compile time.

AI can significantly improve implementation speed, but it does not replace the design phase. Especially for low-level firmware code, the abstractions and invariants should be designed carefully before generating or implementing a large amount of code.

PR #258, opened 2026-08-31
refactor(prototyper): rewrite device drivers in safe Rust over a bounded MMIO abstraction

### LINKED ISSUE #253: Rewrite device drivers in pure safe Rust using an MMIO abstraction
# Rewrite device drivers in pure safe Rust using an MMIO abstraction

Part of #219 (driver layer for `firmware/prototyper`).

## Problem

The device drivers (UART, CLINT, reset, AIA) currently depend on external crates (`uart16550`, `aclint`, `sifive-test-device`, `uart_xilinx`, `bouffalo-hal`, `uart_sifive`, `arm_pl011-uart`, `uart_xscale`) whose types hold raw pointers internally. This forces `unsafe impl Send/Sync` on every wrapper type (19 assertions before #250, 13 in `platform/` alone) and scatters unsafe MMIO dereferences across the codebase.

The wrappers are pass-through adapters with zero logic — they exist solely to bridge external crate types to our `ConsoleDevice`/`IpiDevice`/`ResetDevice` traits. Removing them removes an entire layer of indirection.

## Proposed solution

Write each device driver directly against a safe MMIO handle, eliminating the external crate types and their wrapper layer entirely.

A `Mmio` type represents a bounded device register region: the constructor is `unsafe` (the caller validates the address range from the device tree), and `read`/`write` are safe methods with bounds checking and volatile semantics.

Each driver is a struct holding one `Mmio`, with methods that read and write registers directly:

```rust
pub struct Uart16550 {
    io: Mmio,
}

impl ConsoleDevice for Uart16550 {
    fn write(&self, buf: &[u8]) -> usize {
        for &b in buf {
            while self.io.read::<u8>(LSR) & THRE == 0 {
                core::hint::spin_loop();
            }
            self.io.write(THR, b);
        }
        buf.len()
    }
}
```

No raw pointers, no `unsafe impl Send/Sync`, no external crate dependency — the entire driver is safe Rust.

## Devices to rewrite

| Device | External crate | Wrapper |
|--------|---------------|---------|
| UART 16550 | `uart16550` | `Uart16550Wrap<R>` |
| UART AxiLite | `uart_xilinx` | `UartAxiLiteWrap` |
| UART SiFive | `uart_sifive` | `UartSifiveWrap` |
| UART Bflb | `bouffalo-hal` | `UartBflbWrap` |
| UART PL011 | `arm-pl011-uart` | `UartPl011Wrap` |
| UART Xscale | `uart-xscale` | `UartXscaleWrap` |
| SiFive CLINT | `aclint` | `SifiveClintWrap` |
| THead CLINT | `aclint` | `THeadClintWrap` |
| SiFive Test Device | `sifive-test-device` | `SifiveTestDeviceWrap` |

9 device types, 8 external crates, 9 wrapper structs — all replaced by self-written drivers using `Mmio`.

## Plan

- [ ] Phase 1: UART 16550 + SiFive CLINT + SiFive Test Device (core three, most common)
- [ ] Phase 2: UART variants (AxiLite, SiFive, Bflb, PL011, Xscale)

## Acceptance criteria

- No unsafe code in the device drivers (unsafe lives only inside the `Mmio` type)
- No external device-driver crate dependencies
- Device behavior unchanged: boot logs are byte-identical across the QEMU matrix

Plucky923:
Closes #253.

Replaces the 8 external device-driver crates and their 9 pass-through wrappers
with self-written drivers built directly on a bounded `Mmio` handle:

- `Mmio` validates each access against its FDT-discovered region (checked
  bounds/overflow/alignment); unsafety is confined to this one type
- Console: uart16550, PL011, SiFive, bflb, xscale, AXI-lite
- Interrupt: SiFive/T-Head CLINT and AIA (IMSIC + QEMU APLIC), selected
  explicitly from Devicetree `compatible` strings
- Reset: SiFive test device, T-Head PMIC/I2C
- `riscv-aia`: adds `Mtopei::claim` (AIA v1.0 §3.9) so the IMSIC trap path
  no longer needs hand-written assembly

All `unsafe impl Send/Sync` on driver wrappers are gone; drivers are pure safe
Rust.


library/riscv-aia/src/register/mtopei.rs
```diff
@@ -25,6 +24,28 @@ impl Mtopei {
...
+        // this hart's top machine external interrupt and has no memory effects.
+        unsafe {
+            core::arch::asm!(
+                "csrrw {bits}, 0x35C, zero",
+                bits = out(reg) bits,
+                options(nomem),
+            );
+        }
+        Mtopei::from_bits(bits).iid()
+    }
```
luojia65: Move to module root. In this case caller may use `mtopei::claim()` instead of `mtopei::Mtopei::claim()`, which follows bare-metal Rust register module conventions.

Ref: `riscv::register::mstatus` functions as a reference: https://docs.rs/riscv/latest/riscv/register/mstatus/index.html.

luojia65: If `mtopei::claim()` returns `Mtopei`, it would include `iprio` as well as a convenience. Currently in this commit, we cannot read `mtopei` again after `claim`, as that the CSR read to `mtopei` is special, it has side-effects so that we cannot read it again.

We can implement `claim()` as a wrapper of `read()`.

firmware/prototyper/src/platform/mmio.rs
```diff
@@ -0,0 +1,104 @@
...
+        return None;
+    }
+    let address = base.checked_add(offset)?;
+    (address % align == 0).then_some(address)
+}
+
+/// A bounded handle to one MMIO register block.
+#[derive(Clone, Copy)]
+pub(crate) struct Mmio {
+    base: usize,
```
luojia65, line 47: Only supports physical address of `0 ..= usize::MAX`. RISC-V supports physical address range even wider than virtual address range. For example, in RV32, physical address can be 34-bit wide, so that two `usize`'s should be used to represent a physical address in RISC-V SBI. We may add a note here that existing hardware only uses physical address range that is smaller than or equal to virtual address range, so one `usize` here is sufficient.

firmware/prototyper/src/platform/mod.rs
```diff
@@ -310,76 +339,68 @@ impl BoardInfo {
...
-            self.reset = Some(base_address);
+    fn discover_reset(&mut self, compatible: &str, range: &Range<usize>) {
+        if driver::SIFIVE_TEST_COMPATIBLES.contains(&compatible) {
+            self.reset = Some(range.start);
+            self.record_mmio_range(range);
         }
     }
 
-    /// Discovers the P1 PMIC reset device from a `compatible` match.
     fn discover_pmic_reset(
```
luojia65: This function detects P1 PMIC reset, not reset for all PMICs. There could be PMIC from other brands or vendors installed on RISC-V boards. Rename this functions to represent this is a P1 PMIC.

firmware/prototyper/src/driver/console/axi_lite.rs
```diff
@@ -0,0 +1,70 @@
+//! Xilinx AXI UART Lite.
```
luojia65, line 1: Add reference document link: Xilinx DS741, _LogiCORE IP AXI UART Lite_ (https://docs.amd.com/api/khub/documents/86Ejvsef_6A3PXWuJfH6SA/content)

luojia65 (approved):
Great! It will be even better if following suggestions are applied.

firmware/prototyper/src/driver/reset/pmic.rs
```diff
@@ -0,0 +1,241 @@
+//! SpacemiT P1 PMIC reset through the K1 I2C controller.
```
luojia65: Rename this file from `pmic.rs` to `pmic_spacemit_p1.rs` in case we are going to support other PMICs.

library/riscv-aia/src/peripheral/imsic.rs
```diff
@@ -20,37 +20,25 @@ pub struct Imsic {
...
-/// MSI (Message-Signaled Interrupt) encoding utilities.
-pub mod msi {
-    /// Encodes an interrupt identity into little-endian MSI data.
-    #[inline]
-    pub const fn encode_le(identity: u16) -> u32 {
-        identity as u32
-    }
+/// IMSIC address layout parameters.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct AddressLayout {
```
luojia65: If `AddressLayout` is not a RISC-V AIA standard structure (which means it is implementation or software defined), we should consider move it out and implement in Prototyper. Putting it in `riscv-aia` means this structure exists in hardware, which is not the case in my opinion.

luojia65 (approved):

Issue #190, opened 2026-03-10
[Bug] sbi_pmu_counter_config_matching panics and crashes the system with malformed ecall arguments


bosswnx:
**Description:**
While fuzzing the SBI implementation, I encountered a severe issue where calling the PMU extension's `sbi_pmu_counter_config_matching` function with extreme boundary values causes a complete system crash instead of returning an error code.

**Steps to Reproduce:**
Trigger an `ecall` from S-mode with the following register values:

* **Extension ID (`eid`):** `0x504D55` (PMU)
* **Function ID (`fid`):** `0x2` (`sbi_pmu_counter_config_matching`)
* **`arg0` (`counter_idx_base`):** `0x0`
* **`arg1` (`counter_idx_mask`):** `0xD3D3D300234B40FE`
* **`arg2` (`config_flags`):** `0xD3`
* **`arg3` (`event_idx`):** `0x909000000DFF0000`
* **`arg4` (`event_data`):** `0xAD90909070909090`

**Expected Behavior:**
According to the RISC-V SBI specification, since `event_idx` (arg3) is completely out of the defined event range and `counter_idx_mask` (arg1) contains extremely large out-of-bounds bits, the implementation should cleanly sanitize the input and return an appropriate error code (such as `SBI_ERR_NOT_SUPPORTED` or `SBI_ERR_INVALID_PARAM`) back to S-mode.

**Actual Behavior:**
Instead of handling the invalid parameters gracefully, RustSBI directly panics, resulting in a complete crash of the firmware/system. This lack of boundary sanitization leads to a Denial of Service (DoS) from S-mode.

**Additional Context:**

<img width="1678" height="1646" alt="Image" src="https://github.com/user-attachments/assets/79841ab2-7b72-4dac-830e-c6d615d82f21" />

bosswnx:
I found another bug that will make the hole system crush in the same function. I'll try to fix it.

<img width="1143" height="1608" alt="Image" src="https://github.com/user-attachments/assets/6302f416-d4c4-4d43-ae77-1dde1af47f9e" />

PR #261, opened 2026-09-02
fix(prototyper): reject invalid pmu counter masks

hyy321:
counter_config_matching currently accepts a counter set if the mask contains at least one valid counter. With SKIP_MATCH, a set containing both a configured counter and an out-of-range counter can therefore succeed, even though the SBI specification requires SBI_ERR_INVALID_PARAM when any selected counter is invalid.

Validate the highest selected counter index before matching. Return SBI_ERR_INVALID_PARAM if counter_idx_base + offset overflows or the resulting index is outside 0..total_counters_num.

The regression test combines an already configured counter with `counters_num`, the first invalid index. Before this change, the call returns the configured counter instead of `SBI_ERR_INVALID_PARAM`.

Follow-up to #191. Related to #190.

Tested:

- `cargo fmt --check`
- Clippy for `rustsbi-prototyper` and `rustsbi-test-kernel`
- `cargo prototyper test --timeout 20 --retries 1`
- QEMU boot tests for dynamic and jump modes

### DIFF
--- a/firmware/CHANGELOG.md
+++ b/firmware/CHANGELOG.md
@@ -26,4 +26,7 @@ All notable changes to this project will be documented in this file. See [conven
 - fix(prototyper): temporary PMU fix for possible S-mode DTB modification
 - fix(prototyper): validate DBCN console shared memory range
 
+### Fixed
+- Reject PMU counter masks containing out-of-range counter indices.
+
 ### Removed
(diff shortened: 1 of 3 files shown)


hyy321:
Thanks for the update. I checked the latest main and found that the full counter mask is now validated before matching, which already covers the invalid-counter case this PR was intended to fix.
The test-kernel has also been refactored since this PR was opened, so the implementation part of this PR appears to be superseded by the current code.
Would a dedicated regression test for the SKIP_MATCH case still be useful? If so, I can update this PR to keep only the regression test against the current test structure. Otherwise, I can close it as superseded.

PR #266, opened 2026-09-03
refactor(firmware): introduce runtime memory access

Plucky923:
## Summary

- add a workspace-private firmware Runtime crate with an initial `memory` module
- represent physical addresses and non-empty address ranges explicitly
- register RAM, reserved, and MMIO ranges without installing global state
- provide bounded supervisor-memory access and typed `MmioRegion<T>` handles
- reject conflicting ranges and prevent independent overlapping MMIO acquisition

## Design

Unsafe physical-resource trust is concentrated in `MemoryRegistry::from_ranges`. The returned interfaces keep raw pointers private, require complete in-range supervisor-memory operations, and fix each MMIO window to one access width.

This PR intentionally does not wire Prototyper into the new Runtime. A follow-up can add the Runtime-owned registration seam and migrate one caller at a time.

## Verification

- `cargo test -p rustsbi-firmware-runtime`
- `cargo clippy -p rustsbi-firmware-runtime --all-targets -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p rustsbi-firmware-runtime --no-deps`
- `cargo check -p rustsbi-firmware-runtime --target riscv64imac-unknown-none-elf`
- `cargo check -p rustsbi-firmware-runtime --target riscv32imac-unknown-none-elf`

### DIFF
--- a/firmware/runtime/src/memory/address.rs
+++ b/firmware/runtime/src/memory/address.rs
@@ -1,0 +1,34 @@
+//! Physical address values.
+
+/// An address in the physical address space.
+///
+/// The current representation supports platforms whose physical-address
+/// width does not exceed the native pointer width (`usize`, which equals XLEN
+/// on RISC-V). Wider physical addresses, such as a 34-bit physical address on
+/// RV32, require a different representation and are not supported.
+#[repr(transparent)]
+#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
+pub struct PhysAddr(usize);
+
+impl PhysAddr {
+    /// Creates a physical-address value.
+    #[inline]
+    pub const fn new(value: usize) -> Self {
+        Self(value)
+    }
+
+    /// Returns the address as a native integer.
+    #[inline]
+    pub const fn as_usize(self) -> usize {
+        self.0
+    }
+
+    /// Adds a byte offset without wrapping.
+    #[inline]
+    pub const fn checked_add(self, offset: usize) -> Option<Self> {
+        match self.0.checked_add(offset) {
+            Some(value) => Some(Self(value)),
+            None => None,
+        }
+    }
+}
--- a/firmware/runtime/src/memory/mmio.rs
+++ b/firmware/runtime/src/memory/mmio.rs
@@ -1,0 +1,126 @@
+//! Typed access to memory-mapped device registers.
+
+use core::marker::PhantomData;
+use core::mem::{align_of, size_of};
+
+use super::PhysAddrRange;
+use crate::{Error, Result};
+
+mod sealed {
+    pub trait Sealed {}
+}
+
+/// A fixed-width integer supported by [`MmioRegion`].
+///
+/// This trait is sealed and implemented for `u8`, `u16`, `u32`, and `u64`.
+pub trait MmioValue: sealed::Sealed + Copy {}
+
+macro_rules! impl_mmio_value {
+    ($($value:ty),* $(,)?) => {
+        $(
+            impl sealed::Sealed for $value {}
+            impl MmioValue for $value {}
+        )*
+    };
+}
+
+impl_mmio_value!(u8, u16, u32, u64);
+
+/// A bounded MMIO window accessed in units of `T`.
+///
+/// `T` is fixed when the window is acquired, and values use native byte order.
+/// Mixed-width register blocks use separate, non-overlapping windows.
+pub struct MmioRegion<T: MmioValue> {
+    range: PhysAddrRange,
+    value: PhantomData<T>,
+}
+
+impl<T: MmioValue> MmioRegion<T> {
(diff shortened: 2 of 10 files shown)


firmware/runtime/src/memory/address.rs
```diff
@@ -0,0 +1,29 @@
+//! Physical address values.
+
+/// A physical address without access authority.
+#[repr(transparent)]
+#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
+pub struct PhysAddr(usize);
```
luojia65: Mark that this implemention only supports that size of physical address (PA) <= size of virtual address (VA), as that RISC-V allows that PA > VA in some paging systems. For example, RV32 Sv32 have 34-bit physical address range, which leads to two `usize`s when representing a physical address structure.

We can add a comment here.

firmware/runtime/src/memory/mmio.rs
```diff
@@ -0,0 +1,120 @@
...
+        // 2. `access_address` checked the complete access and its alignment.
+        // 3. Registration keeps the MMIO window accessible.
+        Some(unsafe { (address as *const T).read_volatile() })
+    }
+
+    /// Writes one `T` at byte offset `offset`.
+    ///
+    /// Returns `false` if the access is out of bounds or misaligned. Hardware
+    /// access faults are not recovered.
+    pub fn write(&self, offset: usize, value: T) -> bool {
```
luojia65:
```suggestion
    pub fn write(&self, offset: usize, value: T) -> Result<(), MemoryError> {
```

MemoryError would include `IndexOutOfBounds` (name comes from [here](https://doc.rust-lang.org/stable/core/slice/enum.GetDisjointMutError.html#variant.IndexOutOfBounds)) or `Misaligned`.

firmware/runtime/src/memory/region.rs
```diff
@@ -0,0 +1,106 @@
...
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct PhysAddrRange {
+    start: PhysAddr,
+    end: PhysAddr,
+}
+
+impl PhysAddrRange {
+    /// Creates `[start, end)`, or returns `None` if it is empty or reversed.
+    pub const fn new(start: PhysAddr, end: PhysAddr) -> Option<Self> {
+        if start.as_usize() >= end.as_usize() {
```
luojia65: Define `Ord, PartialOrd` for `PhysAddr`. So we can use:
```suggestion
        if start >= end {
```

firmware/runtime/src/memory/registry.rs
```diff
@@ -0,0 +1,335 @@
...
+        for ram in self.ranges(RangeKind::Ram) {
+            let mut cursor = ram.start().as_usize();
+            for reserved in self.ranges(RangeKind::Reserved) {
+                if !ram.overlaps(reserved) {
+                    continue;
+                }
+                if cursor < reserved.start().as_usize() {
+                    available.push(
+                        PhysAddrRange::new(super::PhysAddr::new(cursor), reserved.start())
+                            .expect("a reserved range follows the RAM cursor"),
```
luojia65: `.expect` means that this program may fail and panic when such condition exists. That will result in a firmware M-mode panic. Try to trap such failures into results and errors.

If we can prove or assure that this exception will definitely not happen (guaranteed by the algorithm) we can add an `assert!` here instead.

firmware/runtime/src/memory/registry.rs
```diff
@@ -0,0 +1,335 @@
...
+    ///
+    /// This succeeds once and returns `None` if no RAM remains.
+    pub fn acquire_supervisor_memory(&mut self) -> Option<SupervisorMemory> {
+        if self.supervisor_memory_acquired {
+            return None;
+        }
+
+        let mut available = Vec::new();
+        for ram in self.ranges(RangeKind::Ram) {
+            let mut cursor = ram.start().as_usize();
```
luojia65: After `Ord, PartialOrd` is implemented, simplify cursor (like `cursor < reserved.start()`).
```suggestion
            let mut cursor = ram.start();
```

firmware/runtime/src/memory/supervisor.rs
```diff
@@ -0,0 +1,131 @@
...
+        };
+        self.regions
+            .iter()
+            .copied()
+            .any(|available| available.contains(range))
+    }
+}
+
+impl ReadMemory for SupervisorMemory {
+    fn read(&self, start: PhysAddr, output: &mut [u8]) -> bool {
```
luojia65:
```suggestion
    fn read(&self, start: PhysAddr, output: &mut [u8]) -> Result<usize, MemoryError>;
```

where 

```
enum MemoryError {
    IndexOutOfBounds,
    // ...
}
```

luojia65 (approved):

PR #268, opened 2026-09-05
refactor(prototyper): manage platform resources through runtime

Plucky923:
## Summary

- derive bounded memory and device resources from the platform description
- bind platform drivers through Runtime-managed MMIO handles
- validate SBI shared-memory access against supervisor memory
- split platform discovery and initialization into focused modules
- claim `mtopei` atomically and preserve QEMU AIA compatibility

## Testing

Passed Library, Prototyper, Docker, openEuler AIA, DCO, and Changelog CI checks.


firmware/runtime/src/spacemit_k1.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    /// Enables the K1 machine-mode cache and prediction features used by firmware.
+    pub fn enable_machine_features(self) {
+        set_csr::<{ K1Csr::MachineSetup as u16 }>(MachineSetup::enabled().0);
+    }
+}
+
+#[cfg(any(target_arch = "riscv32", target_arch = "riscv64"))]
+#[inline]
+fn set_csr<const CSR: u16>(bits: usize) {
```
luojia65, line 136: A generic `set_csr` can't be safe, we can adjust API here in the future.

firmware/prototyper/src/platform/boot.rs
```diff
@@ -1,116 +1,168 @@
...
+            .during("locating the firmware RAM bank")?,
+    );
+
+    let devices = driver::bind_devices(&board, &mut memory).during("binding platform devices")?;
+    let k1_resources = board
+        .spacemit_k1
+        .map(|registers| K1BootResources::acquire(&mut memory, registers))
+        .transpose()
+        .during("acquiring SpacemiT K1 resources")?;
+    let is_k1 = k1_resources.is_some();
```
luojia65, line 60: In following pull requests, we can remove special `is_k1` condition here.

firmware/prototyper/src/platform/info.rs
```diff
@@ -0,0 +1,120 @@
...
+            spacemit_k1: None,
+            hart_count: 0,
+            timebase_frequency_hz: None,
+            enabled_harts: [false; NUM_HART_MAX],
+            model: String::new(),
+            pmic_reset: None,
+        }
+    }
+
+    pub(crate) fn is_qemu_virt(&self) -> bool {
```
luojia65, line 110: We shouldn't judge specially if it is QEMU virt, this should be changed in the future.

firmware/prototyper/src/driver/clint/sifive.rs
```diff
@@ -1,65 +1,158 @@
...
 }
 
-impl TimerDevice for SifiveClint {
+pub(super) fn bind(
+    registers: DeviceRegisterRange,
+    memory: &mut MemoryRegistry,
+) -> runtime::Result<InterruptDevices> {
+    let ipi_window_size = NUM_HART_MAX
+        .checked_mul(size_of::<u32>())
+        .ok_or(runtime::Error::Overflow)?;
```
luojia65, line 67: Sifive CLINT has a maximum number of harts supported. In the case we have more harts than that in the future, we'd check if ipi window size overflows Sifive CLINT IPI register range.

firmware/prototyper/src/driver/console/axi_lite.rs
```diff
@@ -1,59 +1,87 @@
...
+
+struct UartAxiLite {
+    registers: MmioRegion,
 }
 
 impl UartAxiLite {
-    /// Wraps an acquired register block.
-    pub(super) fn new(mmio: Mmio) -> Self {
-        Self { mmio }
+    fn new(registers: MmioRegion) -> Self {
```
luojia65, line 56: Validate MmioRegion width upon struct creation.

firmware/prototyper/src/driver/console/pl011.rs
```diff
@@ -1,83 +1,164 @@
...
+        uart.write_reg(Register::IntBaud, u32::from(divisors.integer));
+        uart.write_reg(Register::FracBaud, u32::from(divisors.fractional));
+        uart.write_reg(
+            Register::LineControl,
+            (LineControl::WORD_LENGTH_8 | LineControl::FIFO_ENABLE).bits(),
+        );
+        uart.write_reg(
+            Register::Control,
+            (Control::UART_ENABLE | Control::TX_ENABLE | Control::RX_ENABLE).bits(),
+        );
```
luojia65, line 124: If the device tree suggests uart pl011 is `okay`, we don't need to configure those registers at all, as previous steps have already configured. We can treat `okay` state as already configured.

If chosen `stdout` in DTB is _not_ `okay`, we assume that this platform is not ready for an SBI firmware, the boot process halts and the user should adjust DTB or former bootloading steps for a working SBI DBCN console.

firmware/prototyper/src/driver/reset/pmic_spacemit_p1/controller.rs
```diff
@@ -0,0 +1,307 @@
...
+            }
+            if status.intersects(mask) {
+                self.clear_status(status);
+                return Some(status);
+            }
+            if self.timebase.elapsed(start, TRANSFER_TIMEOUT) {
+                self.reset();
+                return None;
+            }
+            core::hint::spin_loop();
```
luojia65, line 203: Not empty loop, this line can be removed.
```suggestion
```

firmware/prototyper/src/platform/info.rs
```diff
@@ -0,0 +1,120 @@
...
+    pub(crate) reset: Option<DeviceRegisterRange>,
+    pub(crate) clint: Option<(DeviceRegisterRange, driver::ClintKind)>,
+    pub(crate) imsic: Option<ImsicInfo>,
+    pub(crate) machine_aplic: Option<DeviceRegisterRange>,
+    pub(crate) spacemit_k1: Option<SpacemitK1Registers>,
+    pub(crate) hart_count: usize,
+    pub(crate) timebase_frequency_hz: Option<u32>,
+    pub(crate) enabled_harts: HartEnableList,
+    pub(crate) model: String,
+    pub(crate) pmic_reset: Option<(DeviceRegisterRange, driver::I2cAddress)>,
```
luojia65, line 88:
```suggestion
    pub(crate) spacemit_p1_pmic_reset: Option<(DeviceRegisterRange, driver::I2cAddress)>,
```
Not all PMIC uses I2C, only SpacemiT P1 does this by now. Rename to be clear.

firmware/prototyper/src/sbi/trap/handler.rs
```diff
@@ -210,12 +208,17 @@ pub fn sbi_call_handler(
...
+                }
             }
             legacy::LEGACY_CONSOLE_GETCHAR => {
-                ret.error = console::getchar();
-                ret.value = a1;
+                if let Some(console) = crate::sbi::console() {
+                    ret.error = console.try_read_byte().map_or(usize::MAX, usize::from);
+                    ret.value = a1;
+                }
             }
```
luojia65, line 222: Add a note here suggests:`ret.value` a.k.a. `a1` is not used in legacy extensions, should not be modified, so that `ret.value` restores `a1` after use. `ret.error` is the actual return value for legacy extensions, despite they are named `error`, they are actually `a0` registers. LEGACY_SET_TIMER and LEGACY_CONSOLE_PUTCHAR returns `0` as return value in all conditions. LEGACY_CONSOLE_GETCHAR returns the byte read from console, or `-1` (i.e. `usize::MAX`) if failed.

firmware/prototyper/src/driver/reset/mod.rs
luojia65, line 24:
```
/// Parsed reset type accepted by the SRST driver layer.
///
/// Reserved raw values are intentionally not representable here.
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
pub enum ResetType {
    /// SBI standard reset type 0x00000000.
    Shutdown,
    /// SBI standard reset type 0x00000001.
    ColdReboot,
    /// SBI standard reset type 0x00000002.
    WarmReboot,
    /// Vendor / platform specific reset type: 0xF0000000 ..= 0xFFFFFFFF.
    VendorSpecific(u32),
}

/// Parsed reset reason accepted by the SRST driver layer.
///
/// Reserved raw values are intentionally not representable here.
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
pub enum ResetReason {
    /// SBI standard reset reason 0x00000000.
    NoReason,
    /// SBI standard reset reason 0x00000001.
    SystemFailure,
    /// SBI implementation specific reset reason: 0xE0000000 ..= 0xEFFFFFFF.
    SbiSpecific(u32),
    /// Vendor / platform specific reset reason: 0xF0000000 ..= 0xFFFFFFFF.
    VendorSpecific(u32),
}

/// Fully parsed SRST request.
///
/// This is the unit consumed by the driver layer.
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
pub struct ResetRequest {
    pub reset_type: ResetType,
    pub reset_reason: ResetReason,
}

/// Low-level error category for an SRST backend.
///
/// Important:
/// - `InvalidParam` is intentionally absent.
/// - successful reset is intentionally absent too, because a successful
///   SRST request does not return.
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
pub enum ResetError {
    /// Mapped to SBI RET_ERR_NOT_SUPPORTED.
    NotSupported,
    /// Mapped to SBI RET_ERR_FAILED.
    Failed,
}
// ^^ other types of error cannot be returned by `fn system_reset`.

pub trait ResetBackend {
    /// Check whether this backend implements the given reset request.
    fn supports_reset(&self, req: ResetRequest) -> bool;

    /// Attempt to reset the system.
    ///
    /// Semantics:
    /// - `req` is already parsed and validated by the upper layer;
    /// - `req` is implemented by this backend;
    /// - if this function returns, it must be an error path.
    fn system_reset(&mut self, req: ResetRequest) -> ResetError;
}
```

firmware/prototyper/src/driver/console/mod.rs
```diff
@@ -24,66 +31,52 @@ pub(crate) trait ConsoleDevice: Send {
     fn write(&self, buf: &[u8]) -> usize;
 }
```
luojia65, line 32:
```rust
/// Low-level error category for one backend slice operation.
///
/// Important:
/// - `InvalidParam` is intentionally absent.
///   Physical-memory-range validation and translation belong to the SBI entry layer.
/// - `write_slice` / `read_slice` operate on one concrete contiguous slice only.
///   They are not identical to SBI `write` / `read`.
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
pub enum DbcnError {
    /// Mapped by the SBI adapter layer to SBI RET_ERR_DENIED.
    Denied,
    /// Mapped by the SBI adapter layer to SBI RET_ERR_FAILED.
    Failed,
}

/// A backend that can serve as the byte-oriented data path of an SBI DBCN console.
///
/// Important:
/// - This trait is intentionally *not* named `Uart`.
/// - This trait is also intentionally *not* phrased in terms of SBI calls.
/// - It models one backend operation on one contiguous slice.
///
/// Relationship with SBI DBCN:
/// - SBI `write` may translate one physical memory range into 1..N slices, then
///   call `write_slice` repeatedly.
/// - SBI `read` may translate one physical memory range into 1..N slices, then
///   call `read_slice` repeatedly.
/// - SBI `write_byte` may be implemented by calling `write_slice(&[byte])` once.
pub trait DbcnBackend {
    /// Try to write bytes from one contiguous source slice.
    ///
    /// Returns:
    /// - `Ok(n)` where `0 <= n <= src.len()`, meaning exactly the first `n` bytes
    ///   of `src` are accepted by the backend;
    /// - `Err(DbcnError::Denied)` if writes are not allowed;
    /// - `Err(DbcnError::Failed)` on backend I/O failure.
    ///
    /// This is a non-blocking slice operation. It is not the SBI `write` call itself.
    fn write_slice(&mut self, src: &[u8]) -> Result<usize, DbcnError>;

    /// Try to read bytes into one contiguous destination slice.
    ///
    /// Returns:
    /// - `Ok(n)` where `0 <= n <= dst.len()`, meaning exactly `n` bytes are consumed
    ///   from the backend receive queue and written into the destination slice;
    /// - `Err(DbcnError::Denied)` if reads are not allowed;
    /// - `Err(DbcnError::Failed)` on backend I/O failure.
    ///
    /// This is a non-blocking slice operation. It is not the SBI `read` call itself.
    ///
    /// Note:
    /// this abstraction specifies the backend state transition and the number of
    /// bytes read. It does not model the concrete contents of `dst` yet.
    fn read_slice(&mut self, dst: &mut [u8]) -> Result<usize, DbcnError>;
}
```

firmware/prototyper/src/driver/mod.rs
```diff
@@ -48,12 +74,6 @@ pub(crate) trait IpiDevice: Send {
     }
 }
```
luojia65, line 75:
```
/// One ordinary sPI hart-mask window request forwarded to the backend.
///
/// This backend models only the normal `(hart_mask, hart_mask_base)` window
/// form of the SBI hart-mask encoding.
///
/// The SBI special encoding
/// `hart_mask_base == usize::MAX`
/// means "ignore hart_mask and target all available harts".
/// That special path is intentionally *not* a backend-visible request form.
/// The SBI adaptation layer should expand it into one or more ordinary window
/// requests before calling the backend.
///
/// Design note:
/// - one backend call models one native SBI hart-mask window;
/// - one such window can select at most `XLEN` harts;
/// - this matches common MSWI / CLINT hardware, which is provisioned for the
///   actual hart count rather than for a specification maximum.
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
pub struct IpiRequest {
    pub hart_mask: usize,
    pub hart_mask_base: usize,
}

/// A backend that implements one ordinary backend-side sPI send operation.
///
/// Design note:
/// this trait models the operation after SBI-layer decoding and validation.
/// In particular:
/// - `InvalidParam` is not a backend-visible outcome;
/// - the special SBI "all available harts" encoding is not a backend-visible
///   request form;
/// - if higher layers wish to target all harts, they should split that target
///   set into multiple ordinary `(mask, base)` window requests.
pub trait IpiBackend {
    /// Send supervisor IPIs to the harts selected by one ordinary request window.
    ///
    /// Backend-visible result space:
    /// - `Ok(())`: all request-window target harts were signaled successfully.
    /// - `Err(Failed)`: unspecified or unknown backend failure.
    ///
    /// Requests passed to this function are assumed to have already been:
    /// - decoded from the SBI binary encoding,
    /// - validated by the SBI adaptation layer, and
    /// - normalized so that the SBI special "all available harts" encoding
    ///   has already been expanded away.
    fn send_ipi(&mut self, req: IpiRequest) -> Result<(), IpiError>;
}
```

firmware/prototyper/src/platform/qemu_aplic/mod.rs
```diff
@@ -0,0 +1,67 @@
+//! QEMU `virt` M-level APLIC setup.
```
luojia65, line 1: In the future, we try APLIC in hardware instead of only QEMU.

luojia65 (approved):

luojia65:
Code LGTM, with small improvements that can be applied in future contributions.

PR #276, opened 2026-09-09
refactor(prototyper): introduce fallible driver backends

luojia65:
Those backends matches RISC-V SBI specification and limits what error number this peripheral should return.

For example, this is error type of DBCN extension backend:

```
pub enum DbcnError {
    Denied,
    Failed,
}
```

`InvalidParam` is intentionally absent, for that the DBCN frontend already checks for parameter (physical memory slice) that they are valid. The backend must not return `InvalidParam` error.


zevorn:
@codex review

PR #211, opened 2026-05-10
fix(prototyper): harden hart stacks against silent overflow

Lfan-ke:
## Symptom

Debug-built prototyper panics on QEMU virt + test-kernel:
```
Hart 0 panicked at prototyper/prototyper/src/sbi/trap/boot.rs:83:13:
internal error: entered unreachable code: Boot stage hsm should be start or stop.
```
Release boots cleanly with the same config.

## Root cause

`HartContext` is embedded at the **bottom** of each hart's trap stack. In debug, recursive `serde_device_tree::visit_map` (~1.4-2 KiB per frame) plus `PLATFORM::init` and `patch_device_tree` exceed the 16 KiB stack and silently overwrite `HsmCell.status`. The downstream symptom changes with whichever field gets clobbered:

| stack | failure |
|---|---|
| 16-20 KiB | `boot.rs:83` (HSM corrupted) |
| 24 KiB | `fifo.rs:62` (RFence Fifo corrupted) |
| ≥ 32 KiB | clean boot |

## Fix

1. **`HartContext` out of stack** — `HART_CONTEXTS: [HartContextSlot; NUM_HART_MAX]` static; `ROOT_STACK` now holds raw bytes only. Overflow can no longer silently corrupt hart-local state.
2. **Active overflow detection** — `0x5BA1_BEEF_DEAD_CA5F` canary at every stack bottom plus an `sp` lower-bound check, asserted on every `fast_handler` entry and at the end of `rust_main`. Overflow panics with `hart N stack overflow: ...` instead of corrupting BSS.
3. **Debug-only 64 KiB floor** — `cfg(debug_assertions)` raises `STACK_SIZE_PER_HART` to absorb un-optimised frames. Release and board configs unchanged.

Compile-time `const _: () = assert!(...)` invariants in `cfg.rs` and `trap_stack.rs` lock the floor, the canary's distinguishing pattern, and the `apply_debug_stack_floor` table so future edits cannot silently regress them.

## Verification

QEMU virt smp=2 with the test-kernel:

| build | configured | result |
|---|---|---|
| release | 16 KiB | Base/TIME/sPI/HSM/DBCN/RFNC pass |
| debug | 16 KiB → floored 64 KiB | all SBI tests pass |
| release | 2 KiB (forced overflow) | `hart 0 stack overflow: bottom canary at 0x80041000 was overwritten` |

---

<details>
<summary>📝 <strong>Click for commit rules checklist / 点击查看提交规范检查表</strong></summary>

[...]

PR #279, opened 2026-09-10
fix(prototyper): flatten GPR save in fast_handler to prevent clobbered ctx dereference

### LINKED ISSUE #278: prototyper: fast_handler may dereference a clobbered ctx pointer (a0) when saving trap registers under -O3
### TL;DR (English)

In `firmware/prototyper/src/sbi/trap/mod.rs`, the `save_regs` closure defers saving
the argument registers (`a1..a7`) until deep inside `handle_interrupt` /
`handle_exception`. The trap context pointer (`ctx: FastContext`) only lives in
`a0` at function entry and is **not** preserved by the C ABI across calls, yet
the deferred save block dereferences `a0` as the `TrapHandler` base long after
entry. In optimized builds (`-O3`/LTO) this manifests at runtime: the base
register held a `.rodata` string pointer and the save sequence faulted with a
Store Access Fault **inside the trap handler itself**, which re-enters
`trap_entry`, corrupts the `mscratch`/`sp` swap, and ends in an unrecoverable
double-trap halt. The exact routing by which `a0` gets clobbered before the
save block depends on toolchain version, optimization level and code layout,
and cannot be reliably audited statically — which is precisely why the pattern
must be removed, regardless of whether one blames LLVM or the code pattern.

### 问题描述

当前 `fast_handler` 的实现（PR #174 引入的模式）：

```rust
// Fast path for SBI calls
if let Trap::Exception(Exception::SupervisorEnvCall) = cause {
    return handler::sbi_call_handler(ctx, a1, a2, a3, a4, a5, a6, a7);
}

// Save registers for other traps
let save_regs = |ctx: &mut FastContext| {
    ctx.regs().a = [ctx.a0(), a1, a2, a3, a4, a5, a6, a7];
};

match cause {
    Trap::Interrupt(interrupt) => handle_interrupt(ctx, interrupt, save_regs),
    Trap::Exception(exception) => handle_exception(ctx, exception, save_regs),
}
```

从 RISC-V 调用约定的角度看，这个模式存在一个脆弱的生命周期依赖：

1. `fast_handler` 是 `extern "C"` 函数，`ctx: FastContext`（`#[repr(transparent)]`
   包装的 `TrapHandler` 指针）**只**由 `a0` 传入；`a1..a7` 是陷入现场的参数寄存器。
2. `a0..a7` 都是 caller-saved 寄存器：在 `save_regs` 实际执行之前，任何函数调用
   （`error!` 日志路径、`hart_extension_probe`/`hart_has_extension` 探测、PMU 计数等）
   都**有权**破坏它们。
3. 保存动作被闭包延迟到分发深处，编译器必须在多个基本块之间维持
   "ctx 指针仍然有效"这一假设。在 `-O3`/LTO 下一旦寄存器压力变化，LLVM 可能在
   内联展开的 `save_regs` 中直接复用 `a0` 作为基址而不重新溢出加载。

实际编译产物中可以观察到两个独立事实（以下为某次 release 构建的反汇编摘录，
地址随构建变化）：

其一，函数内存在将 `a0` 改写为其它用途的路径（如以 `a0` 传参的日志调用）：

```asm
xxxx: auipc a0, ...
xxxx: addi  a0, a0, ...      ; a0 <- .rodata 字符串指针
xxxx: jalr  log::log
```

其二，非 ECALL 路径的寄存器保存块距离函数入口很远，且直接以 `a0` 为基址
解引用、未重新加载：

```asm
xxxx: ld   s1, 0(a0)         ; s1 <- 伪 TrapHandler.context
xxxx: ld   s0, 16(a0)
xxxx: sd   s0, 64(s1)
xxxx: sd   a1, 72(s1)        ; 向错误基址写寄存器现场
```

故障现场实测到的事实是：执行到保存块时，基址寄存器中装载的正是 `.rodata`
指针（如 `0x800010c9`），`sd a1, 72(s1)` 写只读段触发 Store Access Fault。
但**污染点与保存块之间的具体 CFG 路由**因工具链版本、优化等级与代码布局
而异，无法静态审计——这正是该模式必须被消除的原因。无论将其归因为 LLVM
的寄存器分配行为还是代码模式缺陷，本质都是该模式把固件最关键路径的正确性
押注在寄存器分配启发式上，而修复方式是同一个：删除这层间接抽象。

值得强调的是，这个闭包间接层**没有任何实际收益**：当前
`handle_interrupt` / `handle_exception` 的每一个 match 分支
（MachineSoft / MachineTimer / MachineExternal、IllegalInstruction、
LoadMisaligned / StoreMisaligned）都无条件调用了 `save_regs`，并不存在
"延迟保存以跳过不必要保存"的收益。删除闭包、在入口平铺保存是严格语义
等价的改写。

另有一个相关但独立的问题是别名安全性：`FastContext` 内部持有
`&'static mut TrapHandler`。若 M-mode 处理函数内的同步异常导致重入，
会有两个 `FastContext` 同时存活并别名同一 `TrapHandler`——这在任何
记忆模型下都是 UB；即便不考虑重入，跨栈保存/恢复的顺序叠加 `&mut` 在
Stacked Borrows 模型下形式上也是 UB（该模型尚未定型）。本修复不试图
解决别名问题本身，但消除"处理函数内部触发同步异常"的诱因后，走入该
UB 区域的路径也随之消失。

### 触发条件与故障表现

- 只有**非 ECALL 陷入**会走上文的慢路径（SBI 调用快速路径不受影响），
  典型即 `MachineTimer` / `MachineSoft` / `MachineExternal` 中断以及
  非对齐访问等异常。
- 是否实际引爆取决于具体构建的代码布局与优化等级：日志串增减、少量代码
  膨胀都可能改变寄存器分配结果，使问题出现或消失，呈现"随机崩溃"表象。
- 一旦引爆：`sd` 写入只读段触发 M-mode 内的 Store Access Fault → 硬件重新
  进入 `trap_entry`，其首条 `csrrw sp, mscratch, sp` 与外层正在使用的栈指针
  发生互换 → 上下文连锁踩踏 → 双重陷阱停机。需要说明的是，进入陷阱处理时
  关闭 MIE 只能掩蔽**中断**；**同步异常不受任何掩蔽**，处理函数内一条
  store fault 即可无条件二次进入 `trap_entry`（本次故障即此路径）。

### 复现与验证方式（无需特殊环境）

**静态验证（确定性的，推荐，两条命令即可完成）**：

1. 以 release 配置构建任意 target 的 prototyper 固件，反汇编并定位
   `fast_handler`：
   ```bash
   rust-objdump -d <构建产物 ELF> | grep -A 200 '<fast_handler'
   ```
2. 检查以下两个特征是否同时成立：(a) 非 ECALL 路径存在形如
   `ld s?, 0(a0)` / `sd a?, 0x48(s?)` 的延迟保存序列；(b) 该序列与函数
   入口之间存在其它调用路径，且 `a0` 在入口之后未被重新建立。两者同时
   成立即为风险构建。

**动态触发（概率性）**：

使用任意会让 S-mode 负载开启时钟中断并进行控制台输出的负载（例如
`sbi-testing` 或自带测试内核），配合上述"风险构建"，当时钟中断恰好在
Guest 输出阶段到达 M-mode 陷阱处理时，可复现 fast_handler 内部的
Store Access Fault 与随后的双重陷阱停机。复现概率与负载的指令数分布、
固件体积相关。

**关于回归测试的说明**：该故障高度依赖代码布局、无法稳定复现，因此
不适合以 CI 回归测试覆盖；上文的静态审计方法即为可落地的验证手段。

### 修复建议

在非 ECALL 分支判定之后、进入任何子调用之前，**立即平铺保存全部参数寄存器**，
并删除 `save_regs` 闭包：

```rust
// 非 SBI 快速路径：入口平铺保存，生命周期即刻终结
let a0 = ctx.a0();
ctx.regs().a = [a0, a1, a2, a3, a4, a5, a6, a7];

match cause {
    Trap::Interrupt(interrupt) => handle_interrupt(ctx, interrupt),
    Trap::Exception(exception) => handle_exception(ctx, exception),
}
```

- SBI 快速路径零开销（不变）；
- 慢路径仅增加 8 条连续 `sd`（约 4~8 周期），且省去闭包间接调用后实际
  开销与原先基本持平；
- 与原实现语义严格等价：原代码所有分支均无条件保存，闭包没有跳过任何
  保存，仅多了一层抽象；
- `handle_interrupt` / `handle_exception` 为 crate 私有函数，签名简化
  不影响任何对外 API。

我已按此方案在本地验证（修改前后反汇编对比确认保存序列紧随入口展开、
不再有跨基本块的寄存器生命周期依赖），可以提交 PR。

x402:
Fixes #278

Complementary to #211 — that PR guards against stack overflow while this one
removes a register-lifetime hazard; the hunks are disjoint, so the two can
land in either order. (Note that #211's entry assertion is itself another
piece of code between trap entry and the deferred save, which further
motivates this fix.) The fast-trap-level follow-up is now tracked in
rustsbi/fast-trap#9.

### Summary

Non-ECALL traps (timer/soft/external interrupts, misaligned accesses, ...)
defer saving the argument registers through a closure
(`save_regs: impl Fn(&mut FastContext)`) that is passed down into
`handle_interrupt` / `handle_exception` and invoked deep inside their
dispatch arms.

This relies on a fragile invariant: `ctx: FastContext` is received **by value
in `a0`** (it transparently wraps the `TrapHandler` pointer), and `a0..a7` are
all caller-saved under the C ABI. Between function entry and the closure body
there are legitimate calls (log macros, `hart_has_extension` probing, PMU
counters) that are free to reuse those registers, while the save block keeps
dereferencing `a0` as the `TrapHandler` base. Under `-O3`/LTO this manifests
at runtime:

```asm
; (addresses from a release build of upstream main @ 26c3a30,
;  rustsbi-prototyper-dynamic.elf; they will differ per toolchain/layout)
8000dea6: mv   a0, a1        ; a0 <- arbitrary call argument
8000deaa: jalr a2            ; an intervening call, free to clobber a0
...
; the deferred save block, ~850 bytes past entry, still dereferences a0 as ctx
8000e088: ld   s1, 0(a0)
8000e08a: ld   s0, 16(a0)
8000e08c: sd   s0, 64(s1)
8000e08e: sd   a1, 72(s1)    ; store to a bogus (read-only) base
```

At runtime the fault was observed with the base register holding a `.rodata`
string pointer while `sd a1, 72(s1)` executed. The exact CFG routing by which
`a0` gets clobbered before reaching the save block depends on toolchain
version, optimization level and code layout, and cannot be reliably audited
statically — which is precisely why the pattern must be removed rather than
patched around. Whatever the attribution (LLVM's allocator or the code
pattern), it bets the correctness of the most critical firmware path on
register-allocation heuristics.

When this fires in M-mode, the store faults (e.g. writing into `.rodata`) and
the CPU re-enters `trap_entry` while the handler is still running. Note that
clearing MIE on trap entry only masks **interrupts**; synchronous exceptions
are masked by nothing, so a single store fault inside the handler
unconditionally re-enters the trap path. The first `csrrw sp, mscratch, sp`
then swaps the live M-mode stack pointer into `mscratch`, and the system ends
in an unrecoverable double-trap halt. Whether a given build is affected
depends on code layout and register allocation, so the failure surfaces as an
apparently random crash that comes and goes with trivial code size changes.

### Fix

Remove the closure entirely. Immediately after the `SupervisorEnvCall`
fast-path check, flatten-save all argument registers into the flow context,
before any subsequent call can touch them:

```diff
     // Fast path for SBI calls
     if let Trap::Exception(Exception::SupervisorEnvCall) = cause {
         return handler::sbi_call_handler(ctx, a1, a2, a3, a4, a5, a6, a7);
     }
 
-    // Save registers for other traps
-    let save_regs = |ctx: &mut FastContext| {
-        ctx.regs().a = [ctx.a0(), a1, a2, a3, a4, a5, a6, a7];
-    };
+    // Flatten-save argument registers for non-SBI traps before dispatching,
+    // so no later call can clobber them or the ctx pointer.
+    let a0 = ctx.a0();
+    ctx.regs().a = [a0, a1, a2, a3, a4, a5, a6, a7];
 
     match cause {
-        Trap::Interrupt(interrupt) => handle_interrupt(ctx, interrupt, save_regs),
-        Trap::Exception(exception) => handle_exception(ctx, exception, save_regs),
+        Trap::Interrupt(interrupt) => handle_interrupt(ctx, interrupt),
+        Trap::Exception(exception) => handle_exception(ctx, exception),
     }
 }
```

All `save_regs(&mut ctx);` calls inside `handle_interrupt` /
`handle_exception` are dropped; their signatures lose the closure parameter.
Both functions are crate-private, so no public API changes.

### Why this shape

- **Fast path untouched**: `SupervisorEnvCall` still returns straight from the
  head of the function; the hottest path (console/IPI/HSM/timer/RFENCE calls)
  gains zero instructions.
- **Slow path cost**: one `ld` + eight consecutive `sd` (~4-8 cycles) per
  non-ECALL trap; dropping the closure indirection means the net cost is
  essentially unchanged.
- **Strictly equivalent rewrite, minus the indirection**: every match arm in
  `handle_interrupt` / `handle_exception` (MachineSoft/Timer/External,
  IllegalInstruction, Load/StoreMisaligned) already calls `save_regs`
  unconditionally — the closure never skipped a single save, so there is no
  "defer to skip unneeded saves" benefit being given up.
- **Deterministic register lifetime**: the save happens before any complex
  control flow, so LLVM no longer needs to track captured caller-saved
  registers across calls and basic blocks. This removes the entire class of
  failure instead of papering over it (stack canaries would only detect the
  corruption after the fact; `#[inline(always)]` on the closure would still
  depend on allocator heuristics).
- After the patch, the save sequence sits right at the top of the non-ECALL
  path in the generated assembly (addresses from the same build):

```asm
8000cf10: ld   s1, 0(a0)       ; ctx is still the pristine entry value here
8000cf12: ld   s0, 16(a0)
8000cf16: sd   s0, 64(s1)
8000cf18: sd   a1, 72(s1)
8000cf1a: sd   a2, 80(s1)
...
```

### Verification

- Before/after disassembly of `fast_handler` on a release build confirms the
  save block moved to the dispatch entry and no longer dereferences `a0` after
  any intervening call.
- With timer interrupts enabled and console output under load, builds that
  previously hit Store Access Faults / double traps inside `fast_handler` run
  cleanly with the patch.
- **No CI regression test is provided on purpose**: the failure is
  code-layout dependent and cannot be reproduced stably. Instead, the risk in
  any build can be audited deterministically with two objdump steps: locate
  `fast_handler` in the disassembly, then check whether a deferred
  `ld s?, 0(a0)` / `sd a?, 0x48(s?)` save sequence survives past intervening
  calls that consume `a0` without it being re-established.

### Note on a fast-trap-level fix

The `fast-trap` crate already hints at this contract in its documentation
("must call `save_args` manually first"), yet the API does not exist. The
follow-up discussion (formal `save_args()` API, and the related
`FastContext(&'static mut TrapHandler)` re-entry aliasing question) is now
tracked in rustsbi/fast-trap#9. This PR is intentionally the zero-dependency
stopgap scoped to prototyper; whether the contract should be formalized at
the fast-trap layer can be discussed there.


x402:
### For discussion context: the flatten-save is the established pattern in this codebase's own lineage (rustsbi-d1)

While preparing the reproducer for #278 I checked how the other RustSBI firmwares structure their trap handling, and found a useful precedent for this patch's shape:

**rustsbi-d1's SEE is the origin of the fast-trap framework used here** (`see/src/trap_vec.rs` starts with `use fast_trap::trap_entry;`), so it shares this exact `fast_handler(ctx, a1..a7)` design. Its non-ECALL branches never defer the register save into a closure — the rdtime emulation branch flattens eagerly at the branch head before any subcall (`see/src/main.rs`, `fast_handler`):

```rust
ctx.regs().a = [ctx.a0(), a1, a2, a3, a4, a5, a6, a7]; // eager, in-branch, before subroutines
```

which is what this PR hoists to the dispatch entry for *all* non-ECALL traps.

The older generation (rustsbi-k210, rustsbi-hifive-unmatched) sits at the other end of the spectrum: `from_machine_save`/`from_supervisor_save` save the full 32-GPR frame in assembly before any Rust runs, so register preservation is structural — immune to this bug class by construction, but paying ~35 stores on every trap, which is what fast-trap's fast path deliberately avoids for ECALLs.

So the two fast-trap rules are complementary rather than in tension:

1. *don't save on the ECALL fast path* — performance (kept as-is here);
2. *once a path needs the argument registers, flatten them immediately* — correctness (what this patch restores, per d1's in-branch pattern).

Also worth noting rustsbi-d1 intercepts mtimer/msoft in vectored naked asm before they ever reach `fast_handler` — I looked at whether prototyper should do the same and would advise against it here: the clear-mtimecmp side effect is platform-specific (CLINT mmio vs Sstc), and an asm bypass would silently skip per-entry invariants such as #211's stack checks. Not proposing it; mentioning only as a datapoint from the same lineage.

Plucky923:
Thanks for the fix and the detailed analysis.

However, the trap handling path in the prototyper has been refactored on `main`, and the `fast_handler` implementation this PR modifies has already been removed. As a result, this PR no longer applies cleanly to the current codebase.

It would be better to first check whether the underlying register-lifetime issue still exists in the new trap handling path. If not, I think we can close this PR (and possibly #278 as well).

x402:
Verified against current `main` (`2ba98ed`): the register-lifetime hazard this PR addressed is gone, and I agree with closing.

What I checked on the refactored path (`firmware/runtime/src/trap/`):

- `fast_handler` and the `fast-trap` dependency are fully gone from the tree — there is no "context over live registers" API anymore, so the deferred-flatten pattern this PR patched no longer exists.
- The naked-asm `trap_entry` (`runtime/src/trap/entry.rs`) saves every caller-saved GPR (`ra`, `gp`, `tp`, `t0`–`t6`, `a0`–`a7`, `t3`–`t6`) plus `mepc`/`mstatus`/`mcause`/`mtval` into a 288-byte `#[repr(C)]` frame on the Runtime stack *before any Rust code runs*.
- Both Rust entry points operate on that stable frame only: `trap_dispatch(frame: &mut TrapFrame)` and the rdtime fast path `try_fast_emulate_time(frame: *mut TrapFrame)` read/write exclusively through the frame pointer. The one remaining deferral — `s0`–`s11` not saved on the fast path — is sound by the Rust ABI (callee-saved registers are preserved by callees), so the bug class #278 described is structurally eliminated.

In other words, the new design combines the structural full-save of the older firmwares (k210/hifive-unmatched) with the fast-path optimization this PR tried to preserve — strictly better than the patch here. I'll close #278 as well; thanks for the quick response!

PR #280, opened 2026-09-14
refactor(firmware): move trap handling and hart lifecycle into Runtime

Plucky923:
Prototyper currently mixes machine trap handling and hart lifecycle mechanisms with SBI extension policy. Move these mechanisms into Runtime, with Prototyper composing platform devices and implementing SBI extensions.

The commits are arranged in dependency order:

1. Add a configurable RustSBI crate path to the derive macro, with documentation and static/dynamic dispatch tests.
2. Add Runtime-owned trap entry, stacks, guarded CSR access, emulation, exception redirection, and hart identity, local storage and lifecycle abstractions.
3. Connect Prototyper through separate IPI, interrupt-controller, timer and trap-accounting interfaces. Keep SBI HSM validation and return-value mapping in Prototyper; drop legacy extension support and remove the old fast-trap implementation and unavailable NACL/SSE adapters.
4. Split the test kernel into focused modules and add trap regression checks.
5. Preserve `mstatus` and declare recovery-clobbered registers across guarded CSR and memory accesses, without changing `recovery_entry`.
6. Reuse the cached Sstc capability for timer calls and skip firmware-counter locking when PMU is unavailable.
7. Name machine-interrupt operations explicitly and leave `stvec`/`sscratch` to supervisor initialization on hart start/resume.

Per-hart capacity and stack sizing now come from Runtime configuration (currently 8 harts and 16 KiB per hart), replacing Prototyper's board-specific settings.

Validation:

- RV32/RV64 Prototyper compilation, including the RV64 hypervisor feature.
- Runtime unit tests: 14 passed. RustSBI integration and xtask tests also passed during initial submission.
- `cargo prototyper test --smp 2 --retries 1` passed.
- QEMU CLINT with Sstc disabled (2 harts) and enabled (4 harts), plus IMSIC with Sstc enabled (4 harts); all include HSM and stopped-hart IPI/RFENCE checks.
- An isolated M-mode QEMU harness checks successful and faulting guarded CSR reads/writes/swaps, including preservation of `mstatus`, `mtvec` and `mscratch`. The pre-fix RV64 implementation fails; the fixed RV32 and RV64 implementations pass.
- CI-equivalent Clippy completed with existing warnings; an additional `-D warnings` check fails on existing lints.
- Intermediate commits checked in an isolated worktree: original Prototyper builds after Runtime is added, and the original test kernel passes after Prototyper switches to Runtime.
- Formatting, whitespace, changelog and commit sign-off checks.

K1 and V861 hardware paths were compile-checked but not tested on hardware.

<details>

</details>


firmware/runtime/src/csr.rs
```diff
@@ -0,0 +1,78 @@
...
+        // SAFETY: Runtime owns the machine trap state on the current hart.
+        unsafe { riscv::register::mie::clear_mtimer() }
+    }
+
+    /// Enables machine external interrupts on the current hart.
+    #[inline]
+    pub fn set_external() {
+        // SAFETY: Runtime owns the machine trap state on the current hart.
+        unsafe { riscv::register::mie::set_mext() }
+    }
```
luojia65: We can theoretically set S-mode interrupt enable/disable bits in `mie`. So in this case, our functions should mark which mode of external, timer, etc. should be used. For example:

```rust
pub fn set_machine_software() { ... }
pub fn clear_machine_software() { ... }
pub fn set_machine_timer() { ... }
pub fn clear_machine_timer() { ... }
pub fn set_machine_external() { ... }
```

firmware/runtime/src/trap/dispatch.rs
```diff
@@ -0,0 +1,296 @@
...
+        // stvec BASE is four-byte aligned; hart entry points may only be
+        // two-byte aligned.
+        if start_addr & 0x3 == 0 {
+            asm!("csrw stvec, {start_addr}", start_addr = in(reg) start_addr, options(nomem));
+        }
+        asm!("csrw sscratch, zero", "csrw sie, zero", options(nomem));
+        sstatus::clear_sie();
+        satp::write(satp::Satp::from_bits(0));
+    }
+}
```
luojia65: RISC-V SBI standard did not define `stvec` initial value for S-mode after SBI HSM reset. In Table 18, Section 9.1, the RISC-V Supervisor Binary Interface Specification version 3.0, [Ref](https://github.com/riscv-non-isa/riscv-sbi-doc/blob/c33ad9f414505806f084e8677e04d2744f76c8df/src/ext-hsm.adoc#function-hart-start-fid-0), it says:

> Request the SBI implementation to start executing the target hart in supervisor-mode, at the address specified by `start_addr`, with the specific register values described in HSM Hart Start Register State.
> 
> HSM Hart Start Register State 
> |Register Name | Register Value |
> |:--|:--|
> |satp          | 0
> |sstatus.SIE   | 0
> |a0            | hartid
> |a1            | `opaque` parameter
> 
> All other registers remain in an undefined state.

It explicitly defined that all other registers including `stvec` and `sscratch` should remain undefined. Removing this procedure conforms the SBI standard.

In the [OpenSBI source code](https://github.com/riscv-software-src/opensbi/blob/3593a5facc4c6938b90429a6973ba9ee21fc5899/lib/sbi/sbi_hart.c#L788-L796) they initialized `stvec` and `sscratch` where I believe it is redundant. Interestingly, I traced these writes all the way back to OpenSBI’s initial commit [here](https://github.com/riscv-software-src/opensbi/blame/4cb4d46875dd0f0dd01d3ff1fee4f0d6cefb857a/lib/sbi_hart.c#L241-L242).

I suggest this fix for this function:

```suggestion
unsafe fn stage_smode_trap_state() {
    unsafe {
        asm!("csrw sie, zero", options(nomem));
        sstatus::clear_sie();
        satp::write(satp::Satp::from_bits(0));
    }
}
```

firmware/runtime/src/trap/dispatch.rs
```diff
@@ -0,0 +1,296 @@
...
+    let param = [
+        frame.read_x(10),
+        frame.read_x(11),
+        frame.read_x(12),
+        frame.read_x(13),
+        frame.read_x(14),
+        frame.read_x(15),
+    ];
+
+    let ret: SbiRet = init::policy().handle_ecall(extension, function, param);
```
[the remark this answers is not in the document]
luojia65: No, this is not necessary. We can just drop those legacy extension support, as they were deprecated dated back to 2019 (7 years ago) in the RISC-V SBI Specification version 0.2. As all mainstream kernels like Linux, Asterinas etc. supports newer SBI DBCN extensions, I believe supporting those legacy extensions is no longer necesary.

Please add 'drop legacy extension support' in either the pull request description or the commit message.

firmware/prototyper/src/sbi/timer.rs
```diff
@@ -8,76 +8,39 @@
...
     /// Sets the timer for the current hart.
     #[inline]
     fn set_timer(&self, stime_value: u64) {
         pmu_firmware_counter_increment(firmware_event::SET_TIMER);
-        let hart_id = current_hartid();
+        let hart_id = HartId::current()
+            .expect("BUG: current hart exceeds Runtime capacity")
+            .as_usize();
 
         if hart_has_extension(hart_id, Extension::Sstc) {
```
luojia65:
```suggestion
        if runtime::trap::has_sstc() {
```

Apply this fix to get 6.5% of performance boost for the `set_timer` SBI call (from 2.806 μs to 2.634 μs on real Avaota F1, Allwinner V821 board hardware). This reuses `has_sstc` cache.

The `set_timer` SBI call is relatively frequent in Linux and other RISC-V kernels.

firmware/prototyper/src/sbi/ipi.rs
```diff
@@ -117,25 +113,29 @@ impl SbiIpi {
...
 
-    /// Clears the specified hart's firmware IPI register.
-    #[inline]
-    pub(crate) fn clear_ipi(&self, hart_id: usize) -> Result<(), IpiError> {
-        self.device.clear_ipi(hart_id)
+impl runtime::ipi::IpiHandler for RuntimeIpiHandler {
+    fn deliver_current(&self) {
+        let ipi_type = get_and_reset_ipi_type();
+        if (ipi_type & IPI_TYPE_SSOFT) != 0 {
+            pmu_firmware_counter_increment(firmware_event::IPI_RECEIVED);
```
luojia65, line 127: This function adds call path for `pmu_firmware_counter_increment`, where it always lock for PMU whether PMU really exists in current platform. Add the following in `pmu_firmware_counter_increment` (**NOT current `deliver_current` function**) for 5.3% of performance speedup of `set_timer` SBI calls.

```rust
    if super::pmu().is_none() {
        return;
    }
```

library/rustsbi/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 
+- Document and test custom RustSBI derive paths for static and dynamic dispatch.
+
```
luojia65:
```suggestion
- Document and test custom RustSBI derive paths for static and dynamic dispatch.
```
Remove the extra blank line.

firmware/runtime/src/trap/recovery.rs
```diff
@@ -0,0 +1,346 @@
...
+            "csrw mstatus, t3",
+            "csrw mscratch, zero",
+            "csrw mtvec, t4",
+            csr = const CSR,
+            in("t3") MPRV_BIT | MXR_BIT,
+            in("t4") prev_mtvec,
+            in("a3") &mut record as *mut RecoveryRecord,
+            inout("t0") data,
+            out("t2") _,
+        );
```
luojia65, line 221:
```suggestion
        asm!(
            "csrw mscratch, a3",
            "lla t2, 2f",
            store_word!(t2 => [a3]),
            // The record is in mscratch; reuse a3 to preserve mstatus across recovery.
            "csrrs a3, mstatus, t3",
            ".option push",
            ".option norvc",
            "2:",
            "csrr t0, {csr}",
            ".option pop",
            "csrw mstatus, a3",
            "csrw mscratch, zero",
            "csrw mtvec, t4",
            csr = const CSR,
            inout("t3") MPRV_BIT | MXR_BIT => _,
            in("t4") prev_mtvec,
            inout("a3") &mut record as *mut RecoveryRecord => _,
            inout("t0") data,
            out("t1") _,
            out("t2") _,
        );
```

firmware/runtime/src/trap/entry.rs
```diff
@@ -0,0 +1,263 @@
...
+        // Allowed-cause match: exceptions only (interrupt bit clear), the
+        // cause's bit must be set in the record's mask.
+        "csrr t1, mcause",
+        "bltz t1, {fail_stop}",
+        "li t2, 32",
+        "bgeu t1, t2, {fail_stop}",
+        "li t2, 1",
+        "sll t3, t2, t1",
+        load_word!(t0[1] => t2),
+        "and t3, t3, t2",
```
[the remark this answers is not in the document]
luojia65: This review can be accepted. In prior `read_csr_guarded` function, `t3` is rewritten by `"csrw mscratch, t3"`. The `t3` in modified in `read_csr_guarded` but only marked `in("t3")` which results in this rewrite. It should be `inout("t3")`.

The same would apply for `t1` where it is modified but we should clobber it so that other code won't use it.

Modify `read_csr_guarded` (**NOT function `recovery_entry`**) as I mentioned in another review comment.

luojia65 (changes requested):
Generally looks okay, but it has performance regressions and some small fixes should be applied.

luojia65 (approved):

PR #282, opened 2026-09-15
ci: close gaps in firmware boot validation

Plucky923:
Fix two gaps in firmware boot validation:

- Add `firmware/runtime/**` to the openEuler AIA pull-request and push path filters so runtime-only changes receive boot coverage.
- Reject expected and forbidden QEMU pattern files that contain only comments or whitespace. These files currently pass the non-empty-file check while silently disabling their corresponding console checks.

Validated locally with controlled QEMU output (reproducing the false pass before the fix), the existing xtask test suite, workflow YAML/configuration checks, `cargo fmt --all --check`, `bash -n`, and `git diff --check`. Live QEMU boot validation is left to CI.


Plucky923:
Checked the final diff: this adds Runtime changes to the openEuler AIA triggers and rejects QEMU pattern files with no effective checks. Firmware code and build behavior are unchanged, and no test files are added. All 21 CI checks passed, including openEuler AIA, Docker and QEMU boot tests. Merging.

PR #284, opened 2026-09-16
fix(prototyper): discover F101 watchdog from device tree

luojia65:
**NOTE: DRAFT PULL REQUEST: MERGE ONLY AFTER https://github.com/YuzukiHD/linux-mainline/pull/1 IS MERGED.** This change depends on upstream device tree change.

Match the D1 watchdog compatible and reuse the V104 reset backend.

Remove the F101 fixed-address fallback, requiring an explicit watchdog node in the device tree.


Plucky923:
The current changes have added a large amount of vendor code to the runtime. I think the adaptation layer may need a better approach to avoid expanding the Runtime as much as possible. We can discuss how to solve it later.

Plucky923 (approved):

Issue #237, opened 2026-08-24
Add SpacemiT K3 platform support


13584452567:
This issue tracks the staged upstreaming of end-to-end SpacemiT K3 support (8× X100 RVA23 cores + 8× A100 AI cores, 60 TOPS, and 2×RT24 cores) to RustSBI, mirroring the vendor OpenSBI implementation ([spacemit_k3.c](https://github.com/spacemit-com/opensbi/blob/k3-br-v1.0.y/platform/generic/spacemit/spacemit_k3.c), branch `k3-br-v1.0.y`, commit `4869910`) and the official [K3 paper](https://forum.spacemit.com/uploads/short-url/60aJ8cYNmrFWqHn4ddwwSzMLjlY.pdf).

The `rebase-boardtest` branch is used as the reference implementation for identifying code that still needs to be split, minimized, reviewed, and submitted upstream. It should not be treated as the upstream merge status itself.

The K3 adds SoC-level requirements beyond the K1 port — A100 cluster parking, PMU power-domain voting, RCPU runtime memory-region protection, and IMSIC interrupt state save/restore on suspend — some of which need new prototyper framework hooks.

- [ ] Refine prototyper framework hooks required by K3
  - [x] Platform-level `emulate_load`/`emulate_store` hook for S-mode MMIO access
  - [ ] `Susp`/`Hsm` suspend platform hooks for IMSIC state save/restore
  - [ ] ~Memory-domain PMP region API (`ENF_PERMISSIONS` semantics)~ OpenSBI specific
  - [x] HSM `hart_start` platform wakeup hook for PMU-based core wakeup
- [ ] Add SpacemiT K3 SoC support to `rustsbi-prototyper` ([PR #257 opened](https://github.com/rustsbi/rustsbi/pull/257); low-level implementation submitted, runtime wiring and hardware validation pending)
  - Status note: #257 contains K3 detection, warmboot/RVBADDR, CCI-550, A100 parking and tuning, PMU/APCR voting, DMASYS setup, S-mode load/store emulation, IMSIC save/restore, and APLIC delegation. These remain unchecked until they are wired into the platform boot, HSM/SUSP, and trap paths and validated on K3 hardware.
  - [ ] K3 platform detection, prioritized over the K1 model fallback
  - [x] C0/C1 warmboot RVBADDR + hardware L2 flush + CCI-550 snoop setup
  - [x] XScale UART console backend for the `spacemit,k3-uart` compatible
[...]

PR #285, opened 2026-09-16
feat(runtime): emulate SpacemiT K1 syscon APMU access faults

hyy321:
## Description

Add load/store access-fault emulation for the SpacemiT K1 syscon APMU, following the virtual-peripheral direction discussed in review.

The changes are:

- Add Runtime load/store access handlers for S-mode access faults. Runtime keeps instruction decoding, register updates and `mepc` advancement.
- Translate the faulting address before dispatch. Bare mode is handled directly and RV64 Sv39 uses a page-table walk.
- Use load/store function pointers instead of the previous `dyn PlatformAccess` interface.
- Discover `spacemit,k1-syscon-apmu` from the device tree and bind it as a Prototyper driver.
- Let the APMU driver own the DT-described MMIO window and define which register sub-ranges may be emulated. The M-mode-only ranges follow the SpacemiT vendor OpenSBI definitions.
- Reuse the APMU device for the existing K1 hart-wakeup path instead of acquiring overlapping MMIO sub-ranges.
- Protect the DT-described APMU window from direct S-mode access with PMP so accesses trap to M-mode.
- Add a Prototyper-side load/store dispatcher between Runtime and the concrete APMU device.
- Preserve the existing trap-redirection path when no access handler is installed or when the driver declines an access.

This only covers the K1 APMU/platform-access part of #237. It does not provide complete K3 platform support or reproduce the full vendor OpenSBI preservation page.

## Performance

Compared no annotation, `#[inline]`, `#[inline(always)]` and `#[inline(never)]` on the access-fault path using QEMU 10.2.4 deterministic instruction counting.

For 256 successful emulated K1 APMU loads:

| Variant | Median guest instructions |
| --- | ---: |
| no annotation | 133895 |
| `#[inline]` | 133895 |
| `#[inline(always)]` | 133895 |
| `#[inline(never)]` | 137479 |

`#[inline(never)]` adds 3584 instructions over 256 accesses, or 14 instructions per emulated load.

`#[inline(always)]` does not reduce the measured instruction count and increases `trap_dispatch` by 52 bytes.

Keep `access_fault` without an explicit inline annotation.

These are QEMU TCG instruction-count results, not K1 hardware latency measurements.

## Validation

- `cargo fmt --all -- --check`
- `git diff --check`
- `cargo check --locked -p rustsbi-prototyper --target riscv64gc-unknown-none-elf`
- `cargo clippy --locked -p rustsbi-prototyper --target riscv64gc-unknown-none-elf -- -D warnings`
- `cargo prototyper test` with QEMU 10.2.4
- Local QEMU fixture covering S-mode load/store → PMP access fault → Runtime dispatcher → K1 APMU driver → M-mode MMIO access → resume

The QEMU fixture uses a DT-described fake APMU region. K1 hardware was not tested.


firmware/runtime/src/trap/dispatch.rs
```diff
@@ -186,6 +178,32 @@ enum Access {
     Store,
 }
 
+/// Load/store access-fault handling: the platform access service may own the
+/// faulting address and complete the instruction in M-mode.
+///
+/// Only a Supervisor-origin fault is offered to the platform; a User-origin
+/// fault keeps the plain redirect, and so does every declined access.
+#[inline(never)]
```
luojia65: Provide performance comparion on which 'inline' should we use. `#[inline(never)]` is maybe not the best choice. We have `#[inline]` or `#[inline(always)]` or no inlines at all in Rust.

firmware/runtime/src/trap/emulate.rs
```diff
@@ -178,6 +178,50 @@ pub fn emulate_store(frame: &mut TrapFrame) -> Result<(), Error> {
...
+    with_trap_facts(|facts| {
+        let (raw, len) = fetch(facts.mepc)?;
+        let op = decode::decode_load(raw)?;
+        let raw_value = platform
+            .emulate_load(facts.mtval, op.kind.width())
+            .ok_or(Error::UnsupportedInstruction)?;
+        frame.write_x(op.rd as usize, op.kind.extend(raw_value));
+        Ok(len)
+    })
+}
```
luojia65: Replace such code with a load/store access dispatcher.

firmware/runtime/src/access.rs
```diff
@@ -0,0 +1,29 @@
+//! Optional platform handling for S-mode load/store access faults.
+//!
+//! Runtime keeps instruction semantics and exposes only the faulting address,
+//! access width, and raw transferred value to the platform.
+
+use spin::Once;
+
+static PLATFORM: Once<Option<&'static dyn PlatformAccess>> = Once::new();
```
luojia65: As a virtual function wrapper, dyn Trait may introduce performance overhead.

luojia65 (changes requested):
The K1/K3 emulation only applies for this special memory range where OpenSBI defined as M-mode only (PMP sections i.e. 'domain' in OpenSBI are applied to guard this pivate memory range). This page can be also visited on Linux, resulting in the firmware have to emulate store or load when reading those memory ranges.

On upstream Linux kernel we may found this out in the device tree [link](https://github.com/torvalds/linux/blob/238650ef6c7c7cca08e032527329424c9fbd70e5/arch/riscv/boot/dts/spacemit/k1.dtsi#L877-L887):

```
syscon_apmu: system-controller@d4282800 {
	compatible = "spacemit,k1-syscon-apmu";
	reg = <0x0 0xd4282800 0x0 0x400>;
	clocks = <&osc_32k>, <&vctcxo_1m>, <&vctcxo_3m>,
		 <&vctcxo_24m>;
	clock-names = "osc", "vctcxo_1m", "vctcxo_3m",
		      "vctcxo_24m";
	#clock-cells = <1>;
	#power-domain-cells = <1>;
	#reset-cells = <1>;
};
```

Here SpacemiT APMU is treated as a peripheral, but is eventually emulated by the SBI firmware. OpenSBI denies the K1/K3 `m_mode_range` when checking if this should be emulated. This is provided as the DTB (as the `a1` opaque register) of SBI firmware input.

In my opinion we may suggest a more neat virtual peripheral architecture for RustSBI. We can support this feature using a _peripheral_ or _device_ support instead:

- Treat 'syscon_apmu' as a peripheral, virtually; its feature being support detailed power management for SpacemiT harts which is at least used by Linux.
- It have a basic address (from DTB, `0xd4282800`) and a memory range (sized `0x400` from the DTB), even if it have no hardware registers on the hardware MMIO interface.
- We virtually support this peripheral through a RustSBI Prototyper driver, like `spacemit_k1_syscon_apmu`.
- This peripheral does not provide standard SBI features (no, the power reset and shutdown is not supported by this virtual peripheral; it has an on-board PMIC to do this). It is used in store/load handler, instead of the ecall handler i.e. SBI dispatcher.
- The driver itself defines what privilege mode can visit what register sub-block, there is already some hardware peripheral that does this (T-Head PLIC).
- So in this way we should introduce a _store/load access fault dispatcher_ for the handler.

The current pull request looks okay but it can be furtherly improved according to the discussion above.

I use the definition of 'virtual peripheral' here for peripheral that is emulated by the supervisor envionment (SEE, as in the RISC-V standard) like firmware etc., that has no hardware interfaces like MMIO. By now no other chips would use such interface in my knowledge, but the framework should be designed carefully to avoid mess in future SBI firmware implementation.

For performance issues, please double check with bench kernel for this case. Do not introduce extra overhead on such execution path which is very frequent on devices that does not support misaligned access. Double check `#[inline(never)]` as it may either improve performace or introduce extra overhead.

luojia65:
For future developers: if you are developing a RISC-V hypervisor that emulates K1/K3 peripherals, and future SpacemiT RISC-V models, this is what you also need to do on the hypervisor host-side.

hyy321:
Thanks for the feedback. I’ll revise the draft in this direction. I’ll replace the current PlatformAccess hook with a load/store access dispatcher, move the K1 syscon APMU handling into the DT-discovered driver path, and remove the explicit #[inline(never)] for now.  
I’ll update the draft after the changes are ready, and add the inline performance comparison separately.

firmware/runtime/src/trap/dispatch.rs
hyy321: I compared no annotation, #[inline], #[inline(always)], and #[inline(never)] with the bench kernel.

With deterministic QEMU instruction counting, no annotation, #[inline], and #[inline(always)] had the same instruction count on this path. #[inline(never)] added 14 guest instructions per emulated load, while #[inline(always)] increased trap_dispatch by 52 bytes without reducing the measured instruction count.

I removed the explicit annotation and kept the default compiler choice.

firmware/runtime/src/trap/emulate.rs
hyy321: I compared no annotation, #[inline], #[inline(always)], and #[inline(never)] with the bench kernel.

With deterministic QEMU instruction counting, no annotation, #[inline], and #[inline(always)] had the same instruction count on this path. #[inline(never)] added 14 guest instructions per emulated load, while #[inline(always)] increased trap_dispatch by 52 bytes without reducing the measured instruction count.

I removed the explicit annotation and kept the default compiler choice.

firmware/runtime/src/access.rs
hyy321: Updated. The dyn PlatformAccess interface has been removed.

Runtime now uses load/store function pointers, while the concrete K1 APMU device stays on the Prototyper side.

firmware/prototyper/src/driver/mod.rs
```diff
@@ -155,6 +158,13 @@ pub(crate) fn bind_devices(
         board.syscon_reboot.filter(|_| sifive_test.is_none()),
         memory,
     )?;
+    // Bind the K1 syscon APMU before K1BootResources so it holds the sole
+    // MmioRegion for that window; K1Wakeup then uses it rather than
+    // acquiring overlapping sub-ranges.
+    let spacemit_k1_apmu = board
```
luojia65, line 164:
```suggestion
    let spacemit_k1_syscon_apmu = board
```

firmware/prototyper/src/driver/spacemit_k1_syscon_apmu.rs
```diff
@@ -0,0 +1,169 @@
+//! SpacemiT K1 syscon APMU (Application PMU) peripheral driver.
+//!
+//! K1 protects parts of its APMU register block from S-mode; an access from
+//! S-mode raises an access fault that M-mode emulates. This driver owns the
+//! whole device-tree-described APMU window and serves both the access
+//! dispatcher and the K1 hart-wake sequence.
+//!
+//! Compatible: "spacemit,k1-syscon-apmu"
```
luojia65, line 8:
```suggestion
```

firmware/prototyper/src/driver/spacemit_k1_syscon_apmu.rs
```diff
@@ -0,0 +1,169 @@
...
+/// The highest wake register end offset: cluster-1 base + 4 entries * 4 bytes.
+const WAKE_MAX_END: usize = WAKEUP_OFFSETS[1] + HARTS_PER_CLUSTER * size_of::<u32>();
+
+/// APMU ranges reachable only from M-mode, as (offset, length) from the APMU
+/// base. An access is declined only when it lies completely inside one range.
+///
+/// Source: SpacemiT vendor OpenSBI `m_only_ranges[]` in
+/// `platform/generic/spacemit/spacemit_k1.c` at commit fc02b891
+/// (spacemit-com/opensbi). Reset-vector ranges (C0_RVBADDR and C1_RVBADDR)
+/// lie outside 0xd4282800..0xd4282c00 and are not included here.
```
luojia65, line 42: Rewrite comments so that we don't show is 'adapted' from OpenSBI. We adapt to spacemit official firmware instead.

firmware/prototyper/src/riscv/spacemit_k1/wakeup.rs
```diff
@@ -9,27 +9,58 @@
...
-    status: MmioRegion,
-    controls: [MmioRegion; 2],
+/// These registers live inside the APMU window. When the device tree
+/// describes the APMU peripheral it owns that window and the wake sequence
+/// goes through it; otherwise the fixed register ranges are acquired here.
+pub(super) enum K1Wakeup {
+    Legacy {
+        status: MmioRegion,
+        controls: [MmioRegion; 2],
+    },
```
luojia65, line 22: Drop `legacy` support if the upstream Linux DTB or that one on real hardware lists K1 specific APMU peripheral.

firmware/runtime/src/trap/emulate.rs
```diff
@@ -178,6 +178,143 @@ pub fn emulate_store(frame: &mut TrapFrame) -> Result<(), Error> {
...
+                0 => (ppn << 12) | offset,
+                1 => ((ppn & !0x1ff) << 12) | (va & ((1 << 21) - 1)),
+                2 => ((ppn & !0x3_ffff) << 12) | (va & ((1 << 30) - 1)),
+                _ => return None,
+            });
+        }
+        table = ppn << 12;
+    }
+    None
+}
```
luojia65, line 316: You don't need to. Use `mstatus::MPRV`.

firmware/runtime/src/access.rs
```diff
@@ -0,0 +1,31 @@
...
+/// The callbacks are indirect calls that avoid a trait object on the
+/// access-fault path.
+pub struct AccessHandlers {
+    /// Emulates a `width`-byte load from physical `addr`, returning the raw
+    /// unsigned value, or `None` when the platform does not own the address.
+    pub load: fn(addr: usize, width: usize) -> Option<usize>,
+    /// Emulates a `width`-byte store of `value` to physical `addr`, returning
+    /// whether the platform owned the address.
+    pub store: fn(addr: usize, width: usize, value: usize) -> bool,
+}
```
luojia65, line 19: Do not add virutal function table here. Use Once of AccessDispatcher. Don't add a new level of abstraction if not necessary.

firmware/prototyper/src/driver/spacemit_k1_syscon_apmu.rs
```diff
@@ -0,0 +1,169 @@
...
+
+/// Publishes the APMU device once during boot.
+pub(crate) fn install(device: SpacemitK1SysconApmu) -> &'static SpacemitK1SysconApmu {
+    APMU.call_once(|| device)
+}
+
+/// Returns the published APMU device.
+pub(crate) fn get() -> Option<&'static SpacemitK1SysconApmu> {
+    APMU.get()
+}
```
luojia65, line 169: This is not necessary. Install into AccessDispatcher.

```suggestion
```

luojia65 (changes requested):
You are adding multiple levels of abstrction and unnecessary `install`'s. Double check your code.

luojia65:
This pull request includes too much change. Split it into smaller PRs.

hyy321:
> This pull request includes too much change. Split it into smaller PRs.

Understood. I’ll split the current changes into smaller PRs, starting with the Runtime-side access-fault dispatcher and MPRV handling, then add the K1 syscon APMU support separately.

hyy321:
> This pull request includes too much change. Split it into smaller PRs.

I checked the existing Runtime path before starting the split. The guarded memory access path already uses MPRV, so I will drop the software Sv39 translation.

One detail I would like to confirm before implementing the first small PR: when you said “Use Once of AccessDispatcher”, do you expect AccessDispatcher itself to live in Runtime or in Prototyper? Prototyper depends on Runtime, but Runtime does not depend on Prototyper, so I want to keep that boundary correct without introducing another callback layer.

luojia65:
@hyy321 use the same mechanism of function `sbi_ecall` in the runtime.

hyy321:
@luojia65 I plan to remove the software Sv39 walker, use the existing MPRV-based access path, and simplify the dispatcher following the same mechanism as runtime `sbi_ecall`.

One point I want to confirm before implementing it: with the current PMP rule denying the DT APMU window to S-mode, an MPRV access with MPP=S would still hit the same PMP restriction after address translation.

Should that PMP rule be adjusted for this emulation path, or is there another intended way to use MPRV here? I want to avoid assuming `mtval` is a physical address or adding another software page-table walk.

luojia65:
@hyy321: You're right: an access with MPRV=1 and MPP=S is still subject to S-mode PMP permissions, so it would fault again on this window. My earlier suggestion to use MPRV was incomplete for this case. I suggest temporarily disable those PMP sections, read with MPRV enabled, then enable those PMP sections back.

Updateing PMP sections require `sfence.vma` instructions according to the RISC-V standard. We should measure the time difference between manual page walking and MPRV on the real board.

hyy321:
I’ve split the generic Runtime access-fault handling into #NEW_PR.

I won’t be continuing the remaining K1-specific changes for now, so I’m closing this PR. Thanks for the review and guidance.

PR #287, opened 2026-09-18
feat(runtime): add access-fault dispatcher

hyy321:
## Summary

This PR extracts the Runtime-side access-fault handling from #285.

It adds an optional `AccessDispatcher` for S-mode load/store access faults. Runtime keeps instruction fetch/decode and register write-back, while the platform dispatcher only handles the actual access.

If no dispatcher is installed, the instruction cannot be decoded, or the dispatcher declines the access, the original access fault is redirected as before.

This PR contains only the generic Runtime infrastructure. The K1-specific syscon APMU, DT, PMP, and MPRV changes from #285 are not included here.

## Validation

- `cargo fmt --all -- --check`
- `cargo check --locked -p rustsbi-firmware-runtime --target riscv64gc-unknown-none-elf`
- `cargo check --locked -p rustsbi-firmware-runtime --target riscv32imac-unknown-none-elf`
- `cargo prototyper build`
- `cargo check --locked -p rustsbi-prototyper --target riscv64gc-unknown-none-elf`
- `cargo clippy --locked -p rustsbi-prototyper --target riscv64gc-unknown-none-elf -- -D warnings`
- `cargo test --locked -p xtask`
- `cargo prototyper test` with QEMU 10.2.4


firmware/runtime/src/trap/dispatch.rs
```diff
@@ -181,6 +173,35 @@ fn misaligned(frame: &mut TrapFrame, access: Access) {
...
 
+/// Load/store access-fault completion through the installed platform
+/// dispatcher.
+///
+/// The fault stays handled in M-mode exactly as before; the dispatcher only
+/// gets the chance to complete the access first, and only for a
+/// Supervisor-origin fault. When it declines — or when none is installed, or
+/// the instruction is not a decodable integer load/store — nothing commits
+/// and the original fault is redirected, which is the behavior of this path
+/// without any dispatcher.
```
luojia65: Make it short and concise. The first paragraph should be only one short sentence, as a short description of the whole function which would appear on rust documentation tools.

```suggestion
/// Complete load/store access faults via the installed platform dispatcher.
///
/// The dispatcher first tries to complete the access, but only for Supervisor-
/// origin faults. If it declines, is absent, or the instruction is not a decodable
/// integer load/store, nothing commits and the original fault is redirected.
```

firmware/runtime/src/trap/emulate.rs
```diff
@@ -178,6 +179,58 @@ pub fn emulate_store(frame: &mut TrapFrame) -> Result<(), Error> {
...
+/// extended value back to `rd`, and advance `mepc`.
+///
+/// The dispatcher performs the access itself, so unlike [`emulate_load`] this
+/// does not touch memory under the trapped context's privilege. `mtval` is
+/// passed on as the faulting address the hardware reported, without further
+/// interpretation.
+///
+/// Returns [`Error::UnsupportedInstruction`] when no dispatcher is installed,
+/// the instruction is not a decodable integer load, or the dispatcher
+/// declines the address; the caller then redirects the original fault.
```
luojia65:
```suggestion
/// Handle a load access fault via the installed [`AccessDispatcher`](super::AccessDispatcher).
///
/// The dispatcher performs the access itself, so unlike [`emulate_load`] it
/// avoids touching memory at the trapped context's privilege. `mtval` is
/// passed through as the hardware-reported faulting address, uninterpreted.
///
/// # Errors
///
/// Returns [`Error::UnsupportedInstruction`] if no dispatcher is installed,
/// the instruction is not a decodable integer load, or the dispatcher declines
/// the address; the caller then redirects the original fault.
```

firmware/runtime/src/trap/emulate.rs
```diff
@@ -178,6 +179,58 @@ pub fn emulate_store(frame: &mut TrapFrame) -> Result<(), Error> {
...
+/// installed [`AccessDispatcher`](super::AccessDispatcher): fetch and decode
+/// the instruction at `mepc`, read `rs2` from the trapped context, offer the
+/// access to the dispatcher, and advance `mepc`.
+///
+/// The dispatcher receives the register value together with the access
+/// width, and performs the width-specific store itself.
+///
+/// Returns [`Error::UnsupportedInstruction`] when no dispatcher is installed,
+/// the instruction is not a decodable integer store, or the dispatcher
+/// declines the address; the caller then redirects the original fault.
```
luojia65:
```suggestion
/// Handle a store access fault via the installed [`AccessDispatcher`](super::AccessDispatcher).
///
/// The dispatcher receives the register value and the access width, and
/// performs the width-specific store itself.
///
/// # Errors
///
/// Returns [`Error::UnsupportedInstruction`] if no dispatcher is installed,
/// the instruction is not a decodable integer store, or the dispatcher
/// declines the address; the caller then redirects the original fault.
```

firmware/runtime/src/trap/frame.rs
```diff
@@ -79,4 +79,11 @@ impl TrapFrame {
         const MPP_MACHINE: usize = 0b11 << 11;
         self.mstatus & MPP_MASK == MPP_MACHINE
     }
+
+    /// Whether the trapped context came from S-mode (`MPP = S`).
+    pub(crate) fn trapped_from_supervisor(&self) -> bool {
```
luojia65: Follow Rust function convention.

```suggestion
    pub(crate) fn is_trapped_from_supervisor(&self) -> bool {
```

firmware/runtime/src/trap/init.rs
```diff
@@ -172,6 +172,50 @@ where
...
+/// hardware refused.
+///
+/// Runtime keeps the instruction semantics: it fetches and decodes the
+/// trapped instruction, extends the loaded value, and advances `mepc`. The
+/// service performs only the access itself, so it receives just the faulting
+/// address reported by `mtval`, the access width in bytes, and — for stores —
+/// the value to write.
+///
+/// Declining (`None`/`false`) leaves the original access fault to be
+/// redirected to the supervisor unchanged.
```
luojia65:
```suggestion
/// A platform service that completes S-mode load/store accesses the hardware
/// refused.
///
/// Runtime keeps the instruction semantics, fetching and decoding the trapped
/// instruction, extending the loaded value, and advancing `mepc`. The service
/// only performs the access itself, so it receives just the `mtval`-reported
/// faulting address, the access width in bytes, and (for stores) the value to
/// write.
///
/// Declining (`Err`) leaves the original fault to be redirected to
/// the supervisor unchanged.
```

firmware/runtime/src/trap/init.rs
```diff
@@ -172,6 +172,50 @@ where
...
+pub trait AccessDispatcher: Sync {
+    /// Completes a load of `width` bytes at `addr`, returning the raw
+    /// unsigned value read, or `None` to decline.
+    fn load(&self, addr: usize, width: usize) -> Option<usize>;
```
luojia65: RISC-V instructions may use different instructions for visiting value kinds even on same address and width, describle it in the function abstraction. For example, the `ld`, `lw` and `lh` visits signed values, while `lwu` and `lhu` visits unsigned ones. We reuse `ValueKind` here.

Consider a new Error type or reusing any error types from this crate here. I list `AccessError` (new) for example.

```suggestion
    fn load(&self, addr: usize, kind: ValueKind) -> Result<usize, AccessError>;
```

firmware/runtime/src/trap/init.rs
```diff
@@ -172,6 +172,50 @@ where
...
+
+    /// Completes a store of the `width` low-order bytes of `value` at `addr`,
+    /// returning `false` to decline.
+    fn store(&self, addr: usize, width: usize, value: usize) -> bool;
```
luojia65:
```suggestion
    fn store(&self, addr: usize, kind: ValueKind, value: usize) -> Result<(), AccessError>;
```

firmware/runtime/src/trap/emulate.rs
```diff
@@ -178,6 +179,58 @@ pub fn emulate_store(frame: &mut TrapFrame) -> Result<(), Error> {
...
+pub fn dispatch_store_fault(frame: &mut TrapFrame) -> Result<(), Error> {
+    reject_machine_origin(frame)?;
+    let dispatcher = init::access_dispatcher().ok_or(Error::UnsupportedInstruction)?;
+    with_trap_facts(|facts| {
+        let (raw, len) = fetch(facts.mepc)?;
+        let op = decode::decode_store(raw)?;
+        let value = frame.read_x(op.rs2 as usize);
+        if !dispatcher.store(facts.mtval, op.kind.width(), value) {
+            return Err(Error::UnsupportedInstruction);
+        }
```
luojia65:
```suggestion
        dispatcher.store(facts.mtval, op.kind, value)?;
```

firmware/runtime/src/trap/emulate.rs
```diff
@@ -178,6 +179,58 @@ pub fn emulate_store(frame: &mut TrapFrame) -> Result<(), Error> {
...
+/// declines the address; the caller then redirects the original fault.
+pub fn dispatch_load_fault(frame: &mut TrapFrame) -> Result<(), Error> {
+    reject_machine_origin(frame)?;
+    let dispatcher = init::access_dispatcher().ok_or(Error::UnsupportedInstruction)?;
+    with_trap_facts(|facts| {
+        let (raw, len) = fetch(facts.mepc)?;
+        let op = decode::decode_load(raw)?;
+        let raw_value = dispatcher
+            .load(facts.mtval, op.kind.width())
+            .ok_or(Error::UnsupportedInstruction)?;
```
luojia65: This may be changed with different error type. I list an example below.

```suggestion
        let raw_value = dispatcher.load(facts.mtval, op.kind)?;
```

firmware/runtime/src/trap/init.rs
```diff
@@ -172,6 +172,50 @@ where
...
+///
+/// Runtime keeps the instruction semantics: it fetches and decodes the
+/// trapped instruction, extends the loaded value, and advances `mepc`. The
+/// service performs only the access itself, so it receives just the faulting
+/// address reported by `mtval`, the access width in bytes, and — for stores —
+/// the value to write.
+///
+/// Declining (`None`/`false`) leaves the original access fault to be
+/// redirected to the supervisor unchanged.
+pub trait AccessDispatcher: Sync {
```
luojia65, line 190: For further development: AccessDispatcher can be nested. A `Once` global dispatcher may be consisted of multiple virtual peripheral driver dispatchers.

luojia65 (changes requested):
Generally good with some modifications.

hyy321:
Thanks for the review. I’ve addressed the requested changes in 49bd732:

- simplified the rustdoc comments and renamed `trapped_from_supervisor` to `is_trapped_from_supervisor`
- changed `AccessDispatcher` to use `ValueKind` and `Result`
- added a minimal `AccessError` and kept dispatcher failures on the existing fault-redirection path
- kept nested dispatcher support out of this PR as suggested for future development

I also re-ran the runtime checks for RV32/RV64, prototyper build/check/clippy, xtask tests, and QEMU prototyper tests.

luojia65 (approved):

PR #288, opened 2026-09-18
refactor(firmware): encapsulate platform-specific devices

Plucky923:
## Summary

This draft rebuilds `refactor/pmp` directly on `main` at `eae4cc7` and replaces the previous 19-commit history with four logically separated, signed-off commits.

- Model V821 and V861 as Runtime SoC capabilities while keeping raw custom-CSR instructions inside architecture-specific implementations.
- Keep the public SoC/CSR types visible on non-RISC-V hosts for Rust Analyzer and static CI; only attempted hardware access is unsupported there.
- Encapsulate reset discovery, priority selection, resource binding, and synchronization inside the reset driver subsystem.
- Separate Allwinner platform preparation, safe firmware drivers, C907 reset entry, and custom SBI adapters.
- Bind V821 custom-extension devices through an opaque Allwinner SBI composition boundary instead of exposing concrete devices through generic boot state.
- Group retained platform facts by memory, hart, device-description, and SoC consumers.

## Commit structure

1. `refactor(runtime): model Allwinner SoC capabilities`
   - Introduces the Runtime SoC capability boundary and removes the old flat Allwinner descriptions.
2. `refactor(prototyper): encapsulate reset driver selection`
   - Keeps reset discovery, binding, backend selection, and synchronization inside the reset subsystem.
3. `refactor(firmware): separate Allwinner driver and SBI layers`
   - Separates safe V821/V861 drivers, the C907 architecture entry, platform preparation, and vendor SBI adapters.
4. `refactor(prototyper): group platform descriptions by consumer`
   - Groups BoardInfo by consumer and completes generic discovery, boot, and publication wiring.

The commits are divided by responsibility; intermediate commits are not intended as standalone build points. The final tree is identical to the previously validated single-commit revision `92d69a2`.

## Validation

- [x] `cargo fmt --all -- --check`
- [x] `git diff --check`
- [x] `cargo test -p rustsbi-firmware-runtime -p rustsbi-prototyper-macros`
- [x] `cargo test -p xtask`
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc -p rustsbi-firmware-runtime --no-deps`
- [x] Runtime checks for `riscv32imac-unknown-none-elf` and `riscv64imac-unknown-none-elf`
- [x] Prototyper checks for `riscv32imac-unknown-none-elf` and `riscv64imac-unknown-none-elf`
- [x] `cargo clippy -p rustsbi-prototyper --target riscv64imac-unknown-none-elf -- -D warnings`
- [x] GitHub Actions: 21/21 checks passed for the reorganized four-commit history

## Local QEMU baseline note

On this host, `cargo prototyper test --timeout 90 --retries 1` boots successfully and passes Base, TIME, IPI, DBCN, RFNC, SRST validation, and misaligned-access tests. It then fails in the test kernel at `firmware/test-kernel/src/trap.rs:170` with `a write to time must redirect to S-mode` (`left: 0`, `right: 2`).

The exact local failure was reproduced from an unmodified detached `main` worktree at `eae4cc7`, so it is an environment/baseline issue rather than a regression from this refactor. The PR's GitHub Actions Prototyper QEMU job passed on the identical final tree and is rerunning for this history update.

The PR remains a draft for architecture and ownership review. A separate custom-extension crate/library split is intentionally left for follow-up; this change establishes the internal boundary first.


firmware/runtime/src/boot/dynamic.rs
```diff
@@ -0,0 +1,73 @@
+//! The fw_dynamic handoff copied at the trusted firmware entry.
```
luojia65:
```suggestion
//! The dynamic firmware (`fw_dynamic`) handoff copied at the trusted firmware entry.
```

firmware/runtime/src/boot/entry.rs
```diff
@@ -0,0 +1,135 @@
...
+    /// Address passed in a1 by the previous stage.
+    pub fn device_tree_address(&self) -> PhysAddr {
+        self.device_tree_address
+    }
+
+    /// Dynamic handoff data copied at entry; static entry supplies none.
+    /// Validation errors are retained for reporting after console initialization.
+    pub fn dynamic_info(&self) -> Option<Result<DynamicInfo, DynamicInfoError>> {
+        self.dynamic_info
+    }
```
luojia65: Rust program does not always provide getters and setters, they prefer to explose the fields when necessary. This structure is a one-time handoff block, modifying the inner fields does not escape from lifetime boundary.

```suggestion
/// Previous-stage data. Only the elected boot hart owns the device-tree grant.
/// This value cannot be constructed or cloned by policy code.
pub struct Handoff {
    /// Address passed in `a1` by the previous stage.
    pub device_tree_address: PhysAddr,
    device_tree: Option<DeviceTreeHandoff>,
    /// Dynamic handoff data copied at entry; static entry supplies none.
    /// Validation errors are retained for reporting after console initialization.
    pub dynamic_info: Option<Result<DynamicInfo, DynamicInfoError>>,
}

impl Handoff {
```

firmware/runtime/src/services.rs
```diff
@@ -0,0 +1,30 @@
...
+pub(crate) struct Services {
+    pub(crate) timer: Option<Option<&'static dyn crate::timer::TimerDevice>>,
+    pub(crate) ipi: Option<Option<&'static dyn crate::ipi::IpiDevice>>,
+    pub(crate) ipi_handler: Option<Option<&'static dyn crate::ipi::IpiHandler>>,
+    pub(crate) external_interrupt: Option<Option<&'static dyn crate::irq::ExternalInterrupt>>,
+    pub(crate) trap_counters: Option<Option<&'static dyn crate::events::TrapCounters>>,
+    #[cfg(any(target_arch = "riscv32", target_arch = "riscv64"))]
+    pub(crate) hart_wakeup: Option<Option<&'static dyn crate::hart::HartWake>>,
+    #[cfg(any(target_arch = "riscv32", target_arch = "riscv64"))]
+    pub(crate) reset_hart: Option<fn(crate::boot::InitialHart)>,
```
luojia65: Those two fields (and its corresponding traits, structs) may always exist under any architecture, to allow static CI checks and Rust Analyzer checks.

firmware/runtime/src/services.rs
```diff
@@ -0,0 +1,28 @@
+//! Immutable startup configuration, published only after boot preparation succeeds.
+
+#[derive(Default)]
+pub(crate) struct Services {
+    pub(crate) timer: Option<Option<&'static dyn crate::timer::TimerDevice>>,
+    pub(crate) ipi: Option<Option<&'static dyn crate::ipi::IpiDevice>>,
+    pub(crate) ipi_handler: Option<Option<&'static dyn crate::ipi::IpiHandler>>,
+    pub(crate) external_interrupt: Option<Option<&'static dyn crate::irq::ExternalInterrupt>>,
+    pub(crate) trap_counters: Option<Option<&'static dyn crate::events::TrapCounters>>,
+    pub(crate) hart_wakeup: Option<Option<&'static dyn crate::hart::HartWake>>,
```
luojia65, line 10: There are nested options. Please double-check.

Plucky923:
@luojia65 Thanks for the review. I have rebuilt this branch on top of the current `main` and narrowed the PR to the platform-specific device refactor, so the earlier boot-handoff and `services.rs` changes referenced by these threads are no longer part of the diff.

The cross-architecture visibility feedback has been carried into the rewritten design: the SoC capabilities, CSR interfaces, and related traits remain available on every architecture for static CI and Rust Analyzer, while only the low-level hardware implementations are architecture-specific. The earlier service representation that introduced nested `Option` values has been removed as well.

I am resolving the now-outdated threads accordingly. Thanks again for pointing these out.

luojia65 (approved):
LGTM

luojia65:
Performance data:

## Bench kernel
SBI Test | Original ns/op | PR #288 ns/op | Change
-- | -- | -- | --
BASE version query | 277.15 | 273.19 | −1.43%
BASE extension probe | 297.98 | 294.86 | −1.05%
mvendorid query | 277.15 | 274.65 | −0.90%
Unsupported extension | 264.65 | 265.69 | **+0.39%**
HSM status query | 329.23 | 328.62 | −0.19%
Set timer | 500.07 | 504.24 | **+0.84%**
Empty IPI request | 382.69 | 377.48 | −1.36%
Empty RFENCE request | 526.99 | 514.49 | −2.37%
Local hart TLB flush | 1043.67 | 1033.40 | −0.98%
Empty DBCN write | 310.48 | 307.35 | −1.01%

## Linux boot test (Allwinner V821)
| Boot metric (average of 5 runs) | Original RustSBI | PR #288  | Difference |
|---|---|---|---|
| Serial first line -> enter shell | 9.7765 s | 9.7664 s | −10.1 ms |
| reboot command -> enter shell | 11.8818 s | 11.8636 s | −18.2 ms |
| RustSBI entry -> kernel handoff | 9.9719 ms | 10.1779 ms | **+0.2060 ms** |

I'll attach performance fixes in follow on review suggestions.

luojia65:
Performance regression is not caused by this pull request. This PR is okay.

Issue #316, opened 2026-09-20
Support Shared SRAM Configuration for SpacemiT K3 AP-RT24 Communication


luojia65:
Part of https://github.com/rustsbi/rustsbi/pull/237.

This adds SpacemiT K3 RT24 core domain support.

- [ ] Add K3 SoC capability detection
  Introduce `SpacemitK3Soc` in Runtime, implement the `Soc` trait, and identify the platform using RV64 and device tree compatibility information.
- [ ] Configure PMA attributes for AP–RT24 shared SRAM
  Locate the PMA entry covering `[0xc0800000, 0xc0880000)`, clean the cache, and set IO attributes while preserving unrelated configuration. Add tests for boundary handling, attribute preservation, and repeated initialization.
- [ ] Integrate shared SRAM PMA initialization into startup
  Retain the K3 SoC capability in Prototyper and invoke Runtime initialization during boot-hart platform preparation and secondary-hart startup.

PR #320, opened 2026-09-20
feat(runtime): add SpacemiT K3 SoC capability detection 

luojia65:
Part of https://github.com/rustsbi/rustsbi/issues/316.

Introduce `SpacemitK3Soc` as a zero-sized capability implementing `Soc`. Require an RV64 target and an enabled device-tree root compatible with `spacemit,k3`.

Prepare the capability for supporting SpacemiT K3 RT24 domain by adding AP-RT24 shared SRAM PMA configuration and its integration into boot-hart and secondary-hart startup.


firmware/runtime/src/soc/spacemit/k3.rs
```diff
@@ -0,0 +1,31 @@
+//! K3 application-processor capability recognized from the Devicetree root.
```
Plucky923, line 1: Could we add a module-level # References section here, as the V821 and V861 SoC modules do? A pinned link to the Linux K3 Devicetree would document the source of the spacemit,k3 compatible and RV64 requirement.

Plucky923 (changes requested):

Plucky923 (approved):
LGTM

PR #318, opened 2026-09-20
ci: boot minimal Linux through RustSBI Prototyper

dongmanyi:
## Summary

Add a separate QEMU CI job for the minimal Linux boot path requested in #314 (part of #307). The job builds RustSBI Prototyper from the PR commit, downloads checksum-pinned Linux 6.6.6 and BusyBox 1.36.1 sources, and builds U-Boot from pinned v2024.04 commit `25049ad560826f7dc1c4740883b0016014a59789`. The kernel embeds a static BusyBox initramfs. Its `/init` checks the RISC-V architecture and `/proc`, prints a distinct userspace marker, and powers off.

The workflow runs on relevant PR changes and by manual dispatch. It fails on a missing marker, boot-error text, nonzero QEMU exit, or timeout, and always uploads the serial log. The existing openEuler AIA workflow is untouched.

## Verification

- Shell syntax and YAML parsing pass locally.
- Archive SHA-256 values come from the Linux kernel and BusyBox release checksum files; the U-Boot archive digest was computed from its pinned GitHub commit archive.
- A full QEMU boot cannot run on this Windows host because QEMU, a RISC-V cross compiler, and Cargo are unavailable. This draft PR is intended to run the Linux CI boot before review.

Refs #314 and #307.

Issue #314, opened 2026-09-20
Add a minimal Linux boot test to RustSBI CI


luojia65:
Part of #307.

Add a QEMU boot test using:

- [x] (Bare) RustSBI Prototyper and a minimal Linux root filesystem. https://github.com/rustsbi/rustsbi/pull/322
- [x] (U-Boot) RustSBI Prototyper -> U-Boot -> Linux kernel and a minimal root filesystem. https://github.com/rustsbi/rustsbi/pull/337
- [x] (EDK II) RustSBI Prototyper -> EDK II -> Linux kernel. https://github.com/rustsbi/rustsbi/pull/336

Verify that init executes a smoke-test command and prints a success marker.

Build RustSBI from the commit under test, pin kernel, root filesystem, and bootloader inputs, fail on boot errors or timeouts, and preserve serial logs.

Reference: [Linux kernel boot guide](https://github.com/rustsbi/rustsbi/blob/main/firmware/docs/booting-linux-kernel-in-qemu-using-uboot-and-rustsbi.md).

zqmz:
Hi! I'd like to work on this. I'll start with the Bare path (RustSBI Prototyper → Linux kernel → minimal initramfs), then add U-Boot and EDK II.

luojia65:
@zqmz The bare SBI path is finished; you can work on U-Boot and EDK II (edk2) paths.

zqmz:
> [@zqmz](https://github.com/zqmz) The bare SBI path is finished; you can work on U-Boot and EDK II (edk2) paths.

I see.I will begin with U-Boot

luojia65:
@dongmanyi You can work with EDK II (edk2) on minimal linux.

dongmanyi:
Thanks, I will take the EDK II path and keep it separate from the U-Boot work. I will build on the merged bare-boot CI, pin the EDK II input, boot its RISC-V QEMU payload on RustSBI Prototyper, reuse the minimal Linux userspace smoke marker and fail-fast checks, and preserve the serial log as an artifact.

luojia65:
Minimum Linux boot is fully supported. Special thanks to: @BattiestStone4, @zqmz and @dongmanyi.

Issue #307, opened 2026-09-20
Add OS boot coverage to RustSBI CI


luojia65:
Add QEMU boot tests for the following operating systems:

- [ ] Arch Linux: Bare, [U-Boot](https://github.com/rustsbi/rustsbi/blob/main/firmware/docs/booting-archlinux-in-qemu-using-uboot-and-rustsbi.md), [EDK2](https://github.com/rustsbi/rustsbi/blob/main/firmware/docs/booting-archlinux-in-qemu-using-edk2-and-rustsbi.md). See #308.
- [x] Fedora: Bare, [U-Boot](https://github.com/rustsbi/rustsbi/blob/main/firmware/docs/booting-fedora-in-qemu-using-uboot-and-rustsbi.md). See #309.
- [ ] FreeBSD: Bare, [U-Boot](https://github.com/rustsbi/rustsbi/blob/main/firmware/docs/booting-freebsd-in-qemu-using-uboot-and-rustsbi.md). See #310.
- [ ] OpenWrt: Bare, [U-Boot](https://github.com/rustsbi/rustsbi/blob/main/firmware/docs/booting-openwrt-in-qemu-using-uboot-and-rustsbi.md). See #311.
- [ ] PolyOS: Bare, [U-Boot](https://github.com/rustsbi/rustsbi/blob/main/firmware/docs/booting-polyos-in-qemu-using-uboot-and-rustsbi.md). See #312.
- [ ] Ubuntu: Bare, [U-Boot](https://github.com/rustsbi/rustsbi/blob/main/firmware/docs/booting-ubuntu-24.04.1-in-qemu-using-uboot-and-rustsbi.md), [EDK2](https://github.com/rustsbi/rustsbi/blob/main/firmware/docs/booting-ubuntu-24.04.3-in-qemu-using-edk2-and-rustsbi.md). See #313.
- [x] Minimal Linux kernel and root filesystem: Bare, [U-Boot](https://github.com/rustsbi/rustsbi/blob/main/firmware/docs/booting-linux-kernel-in-qemu-using-uboot-and-rustsbi.md). See #314.
- [x] openRuyi 2026.08 Zero: QEMU boot coverage added in #317 (merged).
- [ ] eweOS: Bare SBI, U-Boot, or EDK II + Limine. See #321.
- [ ] rCore community kernels (rCore, ArceOS, Starry), on Bare and U-Boot. See #329.
- [ ] Asterinas: on Bare SBI and U-Boot; see #330.
[...]

PR #322, opened 2026-09-20
ci: boot a minimal Linux userspace on Prototyper

BattiestStone4:
## Summary

Adds a CI job that boots a real Linux kernel all the way to userspace through the Prototyper dynamic firmware, for #314.

The firmware is started with `-bios` and the kernel with `-kernel`, so QEMU performs the `fw_dynamic` handoff itself and no bootloader takes part — the approach suggested in review of #318. The boot path is correspondingly short: two builds instead of three, and no `loader` devices or overlapping load addresses to reason about.

Linux and BusyBox come from checksum-pinned release archives, and only the build *products* are cached, under a key pinned to their versions, so a warm run skips both compiles. The initramfs is still rebuilt every run, so edits to `/init` cannot go stale.

`/init` runs under `set -eu` and asserts `uname -m` is `riscv64` and `/proc/version` is readable, so the smoke marker means a riscv64 kernel reached userspace with a mounted `/proc`. The wait loop fails as soon as `Kernel panic`, `not syncing` or `Attempted to kill init` appears, rather than sitting out the boot timeout. Both checks are taken from #318.

## Timing

| run | total | `Boot minimal Linux` step |
| --- | --- | --- |
| cold (first run of a cache key) | 11 min 55 s | 9 min 54 s |
| warm | 2 min | 2 s |

For comparison, this repo's other checks run from 6–12 s (Changelog, DCO) up to ~3.5 min (Prototyper). The cold path only runs when a pinned upstream version changes or the cache is evicted; RustSBI itself is rebuilt from the commit under test every run.

Running it surfaced two toolchain details, both handled in the script and the doc:

- `libc6-dev-riscv64-cross` is installed explicitly. It is only a *recommends* of `gcc-riscv64-linux-gnu`, so `--no-install-recommends` drops it, and BusyBox then fails on a missing `bits/libc-header-start.h`.
- The `tc` applet is disabled: its CBQ constants left the UAPI headers in Linux 6.8, and BusyBox 1.37.0 still has not adapted.

## Verification

Cold run: <https://github.com/rustsbi/rustsbi/actions/runs/35500878373>. Warm run: <https://github.com/rustsbi/rustsbi/actions/runs/35503706756>. Both green, and both serial logs show the handoff and the marker:

```
[RustSBI] INFO - Redirecting hart 0 to 0x00000080200000 in Supervisor mode.
[    0.000000] SBI implementation ID=0x4 Version=0x401
[    0.686181] Run /init as init process
RUSTSBI-SMOKE-OK 6.12.110
[    0.848820] reboot: Power down
```

The whole chain was also run locally in a container matching the `ubuntu-24.04` runner and the same apt list, both as a full cache miss (kernel rebuilt, config assertions exercised) and as a cache hit.

Refs #314, #307.


Plucky923:
Thanks. This overlaps substantially with #318 and implements the same requirement from #314, while incorporating the direct `-bios` + `-kernel` approach suggested during #318's review.

@BattiestStone4 and @dongmanyi, please coordinate on the remaining differences so we can converge on a single implementation rather than maintaining two parallel PRs.

BattiestStone4:
Thanks — I agree we should converge on one implementation, so I've updated this PR to take on the parts of #318 that were better than mine, rather than asking you to pick between two branches.

**Taken from #318** (credited in the commit with `Co-authored-by: dongmanyi`)

- `/init` now runs under `set -eu` and asserts `uname -m` is `riscv64` and `/proc/version` is readable, so the marker means a riscv64 kernel reached userspace with a mounted `/proc`, not just that a shell ran.
- The boot wait loop matches `Kernel panic|not syncing|Attempted to kill init` and fails the moment one appears, printing the matching lines. Before, a panicking kernel sat until the boot timeout and reported only that the marker never appeared.

@dongmanyi if either lands differently from what you intended, say so and I'll adjust — the point is that your checks make it in, not that they get paraphrased.

**On "this is very long"** — that is the fair criticism, so numbers instead of an argument. Recent runs in this repo:

| check | duration |
| --- | --- |
| Changelog, DCO | 6–12 s |
| Library | 1–2 min |
| openRuyi, Docker | 2–2.8 min |
| Prototyper | ~3.5 min |
| Minimal Linux boot (#318) | 10–14 min |
| this job, cold cache | 11 min 55 s |
| this job, warm cache | **2 min** |

The screenshot is #318's job, and that cost repeats on every run: its cache holds the three source archives, so the kernel, BusyBox and U-Boot are all rebuilt each time — there is no run that skips the compiles. This PR caches the build *products* under a key pinned to their exact versions and checksums, so only the first run after a pinned version changes pays for the builds. RustSBI itself is rebuilt from the commit under test every run, always.

Concretely, the `Boot minimal Linux` step went from **9 min 54 s** (cold) to **2 s** (warm, cache restored in 2 s); the job total is now 2 min, below the heaviest existing check. Two structural choices keep the cold path down as well: no U-Boot is built, and the rootfs is passed as an external `-initrd`, so editing `/init` does not invalidate the cached kernel. The timeout is 45 minutes, matching `openruyi.yml` and `openeuler-aia.yml`; 40 s of the remaining 2 min is `cargo install cargo-binutils`, which could be cached too if you want it lower.

**Remaining differences from #318**

#318 boots through U-Boot (`-bios spl/u-boot-spl`, an itb carrying RustSBI, then `booti`); this PR uses the direct `fw_dynamic` handoff, which is the approach suggested in #318's review, and that is what removes the U-Boot build and the `BOOTCOMMAND`/itb configuration. Otherwise:

| | #318 | this PR |
| --- | --- | --- |
| kernel | 6.6.6 | 6.12.110 (current LTS) |
| rootfs | `INITRAMFS_SOURCE`, compiled into the kernel | external `-initrd`, kernel stays cached |
| cache | source archives | build products |
| docs | — | updates `firmware/docs/booting-linux-kernel-in-qemu-using-rustsbi.md` |
| `-cpu rv64,v=true` | needed | not needed: our BusyBox reports `Tag_RISCV_arch: rv64i2p1_m2p0_a2p1_f2p2_d2p2_c2p0_zicsr2p0_zifencei2p0_zmmul1p0` (no V), and the boot is green on QEMU's default CPU |

If #314 is meant to be the U-Boot chain specifically, that is a different test from this one — say so and I'll reshape this PR into it, but the U-Boot build would need caching too or the check stays above ten minutes.

**Verification**

Warm run on this PR, commit `45278ad`: <https://github.com/rustsbi/rustsbi/actions/runs/35503706756> (2 min, all steps green, [log artifact](https://github.com/rustsbi/rustsbi/actions/runs/35503706756/artifacts/10602828494)). Cold run: <https://github.com/rustsbi/rustsbi/actions/runs/35500878373>.

Both serial logs carry the handoff and the userspace marker:

```
[RustSBI] INFO - Redirecting hart 0 to 0x00000080200000 in Supervisor mode.
[    0.000000] SBI implementation ID=0x4 Version=0x401
[    0.686181] Run /init as init process
RUSTSBI-SMOKE-OK 6.12.110
[    0.848820] reboot: Power down
```

The whole chain was also run locally in a container matching the `ubuntu-24.04` runner and the same apt list, both as a full cache miss (kernel rebuilt, config assertions exercised) and as a cache hit.

firmware/docs/booting-linux-kernel-in-qemu-using-rustsbi.md
luojia65, line 1: Out of scope, remove this file for now.

.gitignore
```diff
@@ -15,6 +15,8 @@
 
 qemu-logs/
 /.openeuler-aia/
+/.cache/
+/.minimal-linux/
```
luojia65: Not required for code development. Those CI temporary files will not be uploaded to the repository.

luojia65 (changes requested):

dongmanyi:
Thanks for incorporating the userspace assertions and fail-fast panic detection from #318, and for crediting @dongmanyi as co-author. I agree that this direct `fw_dynamic` boot is the better single implementation for #314. I will close #318 so review can focus here.

I checked this head against the latest `main`: GitHub currently reports a merge conflict confined to `.gitignore`, after `chore: remove local AI tooling from .gitignore` landed. Please resolve that before merge.

.github/scripts/prototyper-minimal-linux-boot.sh
```diff
@@ -0,0 +1,299 @@
...
+      return 0
+    fi
+    if boot_has_failed; then
+      report_boot_failure
+      return 1
+    fi
+    if ! kill -0 "$QEMU_PID" 2>/dev/null; then
+      report_early_exit
+      return 1
+    fi
```
luojia65, line 280: There is a race between checking the success marker and checking whether QEMU is still running: Linux may print the marker and power off between those checks, causing a successful boot to be reported as a failure (CI false negative).

Please recheck the marker after detecting QEMU’s exit and once after the final sleep, before reporting a timeout.

```suggestion
    if ! kill -0 "$QEMU_PID" 2>/dev/null; then
      # QEMU may have printed the marker and powered off since our last read.
      if userspace_is_ready; then
        return 0
      fi
      report_early_exit
      return 1
    fi
```

.github/scripts/prototyper-minimal-linux-boot.sh
```diff
@@ -0,0 +1,299 @@
...
+    sleep 1
+  done
+
+  echo "Linux did not reach userspace within ${BOOT_TIMEOUT_SECS}s" >&2
```
luojia65, line 289:
```suggestion
  # Check once more after the final sleep, including the timeout boundary.
  if userspace_is_ready; then
    return 0
  fi

  echo "Linux did not reach userspace within ${BOOT_TIMEOUT_SECS}s" >&2
```

luojia65 (approved):
Generally okay, with one possible race condition in the script.

luojia65 (approved):

Plucky923 (approved):

luojia65:
Please squash the commits. There are too much commits on this single change.

PR #324, opened 2026-09-20
chore: remove local AI tooling from .gitignore

Plucky923:
Remove `.codex` and `.humanize` from the repository-level `.gitignore`.

`.humanize` is tool-specific local state and does not need to be managed by the RustSBI repository. `.codex`, on the other hand, may contain project-level Codex configuration such as `.codex/config.toml`, which can legitimately be tracked when shared by the project.

Developer-specific local files should be handled by personal/global Git ignore configuration rather than ignored unconditionally at the repository level.

## Testing

- `git diff --check`


luojia65 (approved):
Looks like they are introduced by accident.

Issue #310, opened 2026-09-20
Add FreeBSD boot tests to RustSBI CI


luojia65:
Part of #307.

Add a QEMU boot test for FreeBSD using RustSBI Prototyper and U-Boot. Verify that the kernel mounts the root filesystem and reaches a login prompt.

Build RustSBI from the commit under test, pin guest and bootloader versions, fail on boot errors or timeouts, and preserve serial logs.

Reference: [FreeBSD boot guide](https://github.com/rustsbi/rustsbi/blob/main/firmware/docs/booting-freebsd-in-qemu-using-uboot-and-rustsbi.md).

longneonlight:
may i try on this issue?

luojia65:
@longneonlight: yes sure! I'll assign you to this issue.

PR #335, opened 2026-09-21
ci: add FreeBSD QEMU boot test

longneonlight:
## Summary

Add complete FreeBSD 14.1 QEMU boot coverage for RustSBI Prototyper.

- Build RustSBI Prototyper from the commit under test.
- Test the `sbi` and `u-boot` boot paths in separate matrix jobs.
- Boot FreeBSD 14.1 to a `login:` prompt.
- Fail on fatal boot errors, early QEMU exit, or timeout.
- Use the official FreeBSD archive for the pinned VM image and kernel archive.
- Pin and verify downloaded inputs with SHA-256 digests.
- Use `qemu-system-misc` on Ubuntu 24.04.
- Separate the VM image and kernel caches.
- Upload a separate serial log for each boot path.

## Documentation

- Update the FreeBSD boot guide because the original FreeBSD 14.1 download link is no longer available.

## Validation

- `cargo fmt --check`
- `cargo test --locked -p xtask`
- `cargo prototyper build`
- `bash -n .github/scripts/prototyper-freebsd-boot.sh`
- `shellcheck .github/scripts/prototyper-freebsd-boot.sh`
- `actionlint .github/workflows/freebsd.yml`
- `git diff --check`
- Local bare SBI boot reached a FreeBSD `login:` prompt.
- Local U-Boot boot reached a FreeBSD `login:` prompt.

Refs: #310
Part of #307


luojia65 (commented):
Add a bare SBI bootloading path. By now your code only provides U-Boot bootloading.

longneonlight:
Sorry, I missed the bare SBI bootloading path. I've updated the workflow. Now both bare SBI and U-Boot boot paths are covered.

I noticed that the FreeBSD download link in the documentation is now unavailable, so I updated it to the official archive URL. The guest version remains pinned to FreeBSD 14.1.

luojia65 (approved):
LGTM

